PR #3, opened 2023-10-11
Support annotation + more tests

dstepanov:


sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -56,12 +64,46 @@ public VisitorContext.Language getLanguage() {
     public void write(ClassDef classDef, Writer writer) throws IOException {
         TypeSpec.Builder classBuilder = TypeSpec.classBuilder(classDef.getSimpleName());
         classBuilder.addModifiers(classDef.getModifiersArray());
+        for (PropertyDef property : classDef.getProperties()) {
```
graemerocher, line 67: need a way to generate records too

graemerocher (approved):

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
dstepanov: True, we can mark `ClassDef` as a record but then what should be generated for Kotlin

graemerocher: a kotlin data class

PR #6, opened 2023-10-13
Support generics

dstepanov:


melix (approved):
Maybe we should consider adding tests which actually check the generated sources, because we're relying on "side effects" like the actual compiled code for tests, but not really checking the sources.

dstepanov:
Maybe, but not sure if `contains` checks are better and more reliable

PR #7, opened 2023-10-16
Basic interface generation support

dstepanov:


graemerocher (approved):

melix (approved):
Do I get it correctly that we've moved the cursor a bit, in the sense that now the user isn't supposed to directly generate code using the library they want (e.g javapoet but maybe something else) to generate code via an abstract model? I'm fine with it, but we may end up writing an abstraction layer over javapoet + kotlinpoet in the end.

dstepanov:
That's the idea. I don't think the abstraction will be enough, for complicated cases we would have to fall back to Javapoet to generate methods bodies etc.
And I would like in the future introduce ASM generator to rewrite some of our writers.

PR #10, opened 2023-10-17
Copy JavaPoet

dstepanov:
To support records etc
https://github.com/square/javapoet/pull/981

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,16 +4,26 @@ micronaut-docs = "2.0.0"
 micronaut-test = "3.9.1"
 groovy = "4.0.12"
 spock = "2.3-groovy-4.0"
-managed-javapoet = "1.13.0"
 managed-kotlinpoet = "1.14.2"
 
+google-truth = "1.1.3"
+google-compile-testing = "0.19"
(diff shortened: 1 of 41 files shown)


graemerocher (approved):

melix (commented):
I'm a little uncomfortable at the idea of copying the entire JavaPoet library here, it's going to make maintenance harder. Also if we do so _and_ that we build a wrapper on top of it and KotlinPoet, maybe it's redundant altogether...

dstepanov:
What do you mean by redundant? JavaPoet is not maintained anymore but KotlinPort I think it is

melix:
I mean that if we build something on top of both JavaPoet and KotlinPoet, and that we have to copy the JavaPoet implementation in any case, we might be better building a model that can generate both Java and Kotlin code directly, inspired from these.

dstepanov:
Java and Kotlin poet have different APIs, the other option would be use the Java poet API and convert it to Kotlin, I don’t think that’s a good idea, we don’t really want to use very opinionated Java Poet API

dstepanov:
I did include records support from https://github.com/FabricMC/javapoet repo

PR #23, opened 2024-01-15
Extend the annotations api

andriy-dmytruk:
Extend the annotations api to include all possible annotation definitions


graemerocher (approved):

dstepanov:
Is it compatible with the Kotlin generator?

andriy-dmytruk:
> Is it compatible with the Kotlin generator?

I just spent some time trying to make the same changes for Kotlin, but there are multiple issues.
Mainly, the kotlin poet `AnnotationSpec` does not seem to support having members that are also `AnnotationSpecs`. It seems like with their implementation the member annotations would need to be written to code blocks. Also adding members that are arrays seems to be just as hard, too.

@dstepanov Can you think of a solution?

dstepanov (approved):

andriy-dmytruk:
Thanks

PR #25, opened 2024-01-16
Add support for generating Javadoc

dstepanov:
Not sure how to test it

### DIFF
--- a/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/AnnotationDef.java
+++ b/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/AnnotationDef.java
@@ -127,6 +127,7 @@ public final class AnnotationDef {
                 new VariableDef.StaticField(TypeDef.of(type), "class", TypeDef.of(Class.class))
             );
         } else {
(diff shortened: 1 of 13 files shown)


sourcegen-model/src/main/java/io/micronaut/sourcegen/model/AnnotationDef.java
```diff
@@ -127,6 +127,7 @@ private static Optional<Object> copyAnnotationValue(
                 new VariableDef.StaticField(TypeDef.of(type), "class", TypeDef.of(Class.class))
             );
         } else {
+            System.err.println("BBB " + requiredType + " " + value);
```
graemerocher, line 130: probably a mistake?

graemerocher (changes requested):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/AnnotationDef.java
dstepanov: Removed

graemerocher (approved):

PR #40, opened 2024-01-27
Minor improvements for generated enums and records

altro3:
Enums before:
```java
public enum MyEnum1 {
  A,

  B,

  C;

  public String myName() {
    return this.toString();
  }
}
```

Enums after:
```java
public enum MyEnum1 {

  A,
  B,
  C;

  public String myName(
  ) {
    return this.toString();
  }
}
```

Records before:
```java
@Builder
public final record MyRecord1(List<String> copyAddresses, @Deprecated int id, String name, @Deprecated(since = "xyz", forRemoval = true) Integer age,
    List<String> addresses, List<?> tags) {
  @Vetoed
  public static MyRecord1Builder builder() {
    return MyRecord1Builder.builder();
  }
}
```

Records after:
```java
@Builder
public final record MyRecord1(
  List<String> copyAddresses,
  @Deprecated int id,
  String name,
  @Deprecated(since = "xyz", forRemoval = true) Integer age,
  List<String> addresses,
  List<?> tags
) {
  @Vetoed
  public static MyRecord1Builder builder(
  ) {
    return MyRecord1Builder.builder();
  }
}
```


graemerocher:
Thanks for the contribution, there are test failures for this one.

altro3:
@graemerocher fixed

PR #54, opened 2024-02-07
Fix method parameters rendering

andriy-dmytruk:
The current code causes a stack overflow exception as it tries rendering the original expression instead of each parameter.

@dstepanov Please suggest where I can create a test for this if a test is required.


dstepanov:
Maybe you can create new generate bean test?

andriy-dmytruk:
@dstepanov Is the test fine?

dstepanov (approved):

PR #72, opened 2024-03-19
Extend JavaPoetSourceGenerator.renderExpression to String[]

rcrida:
Expand JavaPoetSourceGenerator.renderExpression to accept a TypeDef.Array when values are String[].

### DIFF
--- a/sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
+++ b/sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
@@ -49,7 +49,9 @@ import io.micronaut.sourcegen.model.StatementDef;
 import io.micronaut.sourcegen.model.TypeDef;
 import io.micronaut.sourcegen.model.VariableDef;
 
+import java.lang.reflect.Array;
 import java.util.Arrays;
+import java.util.stream.IntStream;
 import javax.lang.model.element.Modifier;
 import java.io.IOException;
 import java.io.Writer;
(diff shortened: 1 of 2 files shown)


sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -422,6 +423,17 @@ private CodeBlock renderExpression(@Nullable ObjectDef objectDef, MethodDef meth
                     case "double" -> CodeBlock.of(constant.value() + "d");
                     default -> CodeBlock.of("$L", constant.value());
                 };
+            } else if (type instanceof TypeDef.Array array) {
+                if (array.componentType() instanceof ClassTypeDef arrayClassTypeDef &&
+                        constant.value() instanceof String[] stringArray) {
+                    final var values = Arrays.stream(stringArray)
+                            .map(string -> CodeBlock.of("$S", string))
```
andriy-dmytruk, line 430: It would be good perhaps to move the constant to a new method. Something like:
```java
         private CodeBlock renderConstantExpression(@Nullable TypeDef type, Object value) {
            if (type instanceof TypeDef.Primitive primitive) {
                return switch (primitive.name()) {
                    case "long" -> CodeBlock.of(value + "l");
                    case "float" -> CodeBlock.of(value + "f");
                    case "double" -> CodeBlock.of(value + "d");
                    default -> CodeBlock.of("$L", value);
                };
            } else if (type instanceof TypeDef.Array array) {
                 final var values = Arrays.stream(array)
                         .map(value -> renderConstantExpression(array.componentType(), value))
                         .collect(CodeBlock.joining(", "));
                 return CodeBlock.concat(
                         CodeBlock.of("new $N[] {", arrayClassTypeDef.getSimpleName()),
                         values,
                         CodeBlock.of("}"));
            } else if (type instanceof ClassTypeDef classTypeDef) {
                String name = classTypeDef.getName();
                if (ClassUtils.isJavaLangType(name)) {
                    return switch (name) {
                        case "java.lang.Long" -> CodeBlock.of(constant.value() + "l");
                        case "java.lang.Float" -> CodeBlock.of(constant.value() + "f");
                        case "java.lang.Double" -> CodeBlock.of(constant.value() + "d");
                        case "java.lang.String" -> CodeBlock.of("$S", constant.value());
                        default -> CodeBlock.of("$L", constant.value());
                    };
                } else {
                    return CodeBlock.of("$L", constant.value());
                }
            }
```
Then by recursively calling it, we could render any array instead of only string arrays.
I didn't try if this will work, though.

rcrida: Good idea, I contemplated doing something recursive, but I should have followed through...

graemerocher (approved):

Issue #9, opened 2023-10-17
Add documentation


graemerocher:
### Feature description

We will need some documentation for this project

PR #95, opened 2024-05-21
document how to write a source generator

graemerocher:
Fixes https://github.com/micronaut-projects/micronaut-sourcegen/issues/9


src/main/docs/guide/generators.adoc
```diff
@@ -0,0 +1,25 @@
+To write a source generator you have a few options. If you only wish to support Java you can use the https://github.com/square/javapoet[JavaPoet API] that is forked into the `io.micronaut.sourcegen.javapoet` package. This fork includes support for Java 17+ constructors like Records.
```
dstepanov, line 1: Can you add a note that JavaPoet is no longer being developed? That's why it's forked

PR #113, opened 2024-07-05
Add `@Singular` + Switch statement/expression, arrays, better API

dstepanov:


sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/BuilderAnnotationVisitor.java
```diff
@@ -136,70 +120,401 @@ public void visitClass(ClassElement element, VisitorContext context) {
...
+        return fieldDef.build();
+    }
+
+    private static void createSingularPropertyMethods(ClassDef.ClassDefBuilder classBuilder,
+                                                      PropertyElement beanProperty,
+                                                      Function<VariableDef.This, StatementDef> returningExpressionProvider) {
+        String propertyName = beanProperty.getSimpleName();
+        String singularName = beanProperty.stringValue(Singular.class).orElse(null);
+        if (singularName == null) {
+            if (propertyName.endsWith("s")) {
```
graemerocher: not sure if we should make this logic smarter. Lombok seems to have stripped the logic from Rails to do this https://github.com/projectlombok/lombok/blob/master/src/core/lombok/core/handlers/singulars.txt

See https://github.com/projectlombok/lombok/blob/master/src/core/lombok/core/handlers/Singulars.java

dstepanov: Added

graemerocher (approved):
Ok looks good. Next we need to add some tests to make sure Micronaut Serialization plus `@Builder` / `@SuperBuilder` / `@Singular` works automatically especially with types that have `private` constructors. 

We should automatically be applying https://github.com/micronaut-projects/micronaut-serialization/blob/6b962e4f964a3a37f622f5f3e620b3abfc51498d/serde-jackson-tck/src/main/java/io/micronaut/serde/jackson/builder/TestBuildMe2.java#L6-L9

dstepanov:
Should it be by default? Maybe it should be implemented in serde.

We might want to add some hint like `@Buildable(MyBeanBuilder.class)` and annotate the original class with the builder class reference. Serde can implement annotating it with `@JsonDeserialize`

graemerocher:
yeah probably makes sense to make it opt in though I wonder if the original class' constructor is private we would need to use the builder anyway?

dstepanov:
Yeah but I'm thinking where to implement it, I think it should be part of Serde to recognize the sourcegen builder annotation and apply some logic

PR #123, opened 2024-07-22
Track processed classes. Fixes #122

### LINKED ISSUE #122: Compilation error with @Wither - Output stream or writer has already been opened
### Expected Behavior

Use Wither without Build failing

### Actual Behaviour

I'm was working only with @Builder without any problem, I;ve tried to work with @Wither to avoid copy records and got error compilation. 

> Task :compileJava FAILED
1 actionable task: 1 executed

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileJava'.
**> java.lang.RuntimeException: java.io.IOException: Output stream or writer has already been opened.**

The error occurr when implement the interface *Wither and try to build

### Steps To Reproduce

1. Create new micronaut project with Gradle (Kotlin)
2. Add      annotationProcessor("io.micronaut.sourcegen:micronaut-sourcegen-generator-java")
    implementation("io.micronaut.sourcegen:micronaut-sourcegen-annotations")
3. Build
4. Create a record and add @Wither
5. Build
6. Implement the interface *Wither
7. Build and Error

### Environment Information

Operating System: Windows 11
Micronaut Version: 4.5.0
JDK Version: 21
Gradle 8.8

### Example Application

_No response_

### Version

4.5.0

graemerocher:
Second round of compilation results in visiting the same class again. This prevents that.

### DIFF
--- a/sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/BuilderAnnotationVisitor.java
+++ b/sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/BuilderAnnotationVisitor.java
@@ -38,6 +38,8 @@ import io.micronaut.sourcegen.model.StatementDef;
 import io.micronaut.sourcegen.model.TypeDef;
 import io.micronaut.sourcegen.model.VariableDef;
 
+import java.util.HashSet;
+import java.util.Set;
 import javax.lang.model.element.Modifier;
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


graemerocher:
@dstepanov not sure if this is the correct fix and it will probably need to be applied to the SuperBuilderVisitor. Currently the same record is visited again in the second round of compilation.

dstepanov:
I think this might be a Core problem

graemerocher:
Agreed, we probably need a fix for existing versions of Micronaut though

dstepanov (approved):

PR #135, opened 2024-08-20
Add an ability to introspect builders

andriy-dmytruk:
Adds the ability to introspect builders with a configuration in the annotations:`@Builder(intropsected=)` and `@SuperBuilder(introspected=)`. By default it is set to true.

### DIFF
--- a/test-suite-java/src/test/java/io/micronaut/sourcegen/example/AnimalSuperBuilderTest.java
+++ b/test-suite-java/src/test/java/io/micronaut/sourcegen/example/AnimalSuperBuilderTest.java
@@ -16,9 +16,12 @@
 package io.micronaut.sourcegen.example;
 
 import java.lang.reflect.Modifier;
+
+import io.micronaut.core.beans.BeanIntrospection;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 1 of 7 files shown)


test-suite-java/src/test/java/io/micronaut/sourcegen/example/AnimalSuperBuilderTest.java
```diff
@@ -61,6 +64,13 @@ public void testDog() {
     }
 //end::test[]
 
+    @Test
+    public void dogIntrospection() {
+        var introspection = BeanIntrospection.getIntrospection(DogSuperBuilder.class);
+        assertNotNull(introspection);
+        assertEquals(6, introspection.getBeanProperties().size());
```
andriy-dmytruk: @graemerocher  I could not make introspection work for the `@SuperBuilder` case.

For `@Builder` I just added `@Creator` on the PersonBuilder all-args contructor which creates a correct introspection. Here there is no all-args constructor, so I tried `@Introspected(withPrefix = "")` on `DogBuilder`, but that does not help resolving the properties.

Do you have any ideas?

Note, that I want to configure introspection for builder separately since this does not work:
```
@Introspected(builder = @IntrospectionBuilder(
    builderClass = DogSuperBuilder.class
))
public class Dog {
```
```
Builder class not found on compilation classpath: <error>
```

andriy-dmytruk: Actually, solved it by adding an all-args constructor for the `SuperBuilder`, too

graemerocher (approved):

Issue #126, opened 2024-08-07
Generate toString, equals and hashCode implementations


andriy-dmytruk:
### Feature description

We can generate implementations for these methods to simplify development. We should probably include superclasses.

Since we cannot generate methods in the original class, we could generate something like `[classname]Utils` similarly to the way we generate `[classname]Builder`. Then the original class would reference the generated utility methods instead of needing to implement them similarly to how `builder()` static method can return a new builder.

Swagger generator mustache templates can be used as a reference, although we probably cannot support all the features, like hiding password in `toString`: 
```java
@Override
    public String toString() { {{#format-2}}
        java.lang.StringBuilder sb = new java.lang.StringBuilder();
        sb.append("{{classname}} (");
        {{#allVars}}
        sb.append("{{^-first}}, {{/-first}}{{name}}=");
            {{^isByteArray}}
                {{#isFormatPassword}}
        sb.append({{#isFormatPassword}}"<redacted>"{{/isFormatPassword}});
                {{/isFormatPassword}}
                {{^isFormatPassword}}
        sb.append(String.valueOf({{getter}}()));
                {{/isFormatPassword}}
            {{/isByteArray}}
            {{#isByteArray}}
        sb.append(String.valueOf({{getter}}()) + ({{getter}}() != null ? " (byte[" + {{getter}}().length + "])" : ""));
            {{/isByteArray}}
        {{/allVars}}
        sb.append(")");
        return sb.toString();
        {{/format-2}}
    }

    @Override
    public boolean equals(Object o) {
        if (this == o) {
            return true;
        }
        if (!(o instanceof {{classname}})) {
            return false;
        }

        {{classname}} other = ({{classname}}) o;
        return {{#vars}}{{^-first}}    {{/-first}}{{^isByteArray}}java.util.Objects.equals(this.{{name}}, other.{{name}}){{/isByteArray}}{{#isByteArray}}java.util.Arrays.equals(this.{{name}}, other.{{name}}){{/isByteArray}}{{^-last}} &&
        {{/-last}}{{/vars}}{{^parent}}{{/parent}}{{#parent}}{{#hasVars}} &&
            {{/hasVars}}super.equals(other){{/parent}}{{^parent}}{{^hasVars}}true{{/hasVars}}{{/parent}};
    }

    @Override
    public int hashCode() {
        final int PRIME = 59;
[...]

PR #146, opened 2024-09-22
Add support for casting expressions

andriy-dmytruk:
Add support for casting expressions that will look like:
* `(Type) value` for Java/Groovy
* `value as Type` for Kotlin

Seems required for https://github.com/micronaut-projects/micronaut-sourcegen/issues/126.

### DIFF
--- a/test-suite-custom-generators/src/main/java/io/micronaut/sourcegen/custom/visitor/GenerateMyBean3Visitor.java
+++ b/test-suite-custom-generators/src/main/java/io/micronaut/sourcegen/custom/visitor/GenerateMyBean3Visitor.java
@@ -22,24 +22,20 @@ import io.micronaut.inject.ast.ClassElement;
 import io.micronaut.inject.processing.ProcessingException;
 import io.micronaut.inject.visitor.TypeElementVisitor;
 import io.micronaut.inject.visitor.VisitorContext;
-import io.micronaut.sourcegen.custom.example.GenerateMyBean1;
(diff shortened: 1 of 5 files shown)


graemerocher (approved):

test-suite-custom-generators/src/main/java/io/micronaut/sourcegen/custom/visitor/GenerateMyBean3Visitor.java
```diff
@@ -72,6 +68,20 @@ public void visitClass(ClassElement element, VisitorContext context) {
                 ))
                 .build()
             )
+            .addMethod(MethodDef.builder("castPrimitive")
+                .addParameter(ParameterDef.of("value", TypeDef.primitive(Double.TYPE)))
+                .returns(TypeDef.primitive(Float.TYPE))
+                .addStatement(new StatementDef.Return(
+                    new ExpressionDef.Cast(
```
dstepanov, line 75: Please add a default method to `ExpressionDef` `cast`, so this statement can we written as:
```
        new VariableDef.MethodParameter(
            "value",
            TypeDef.primitive(Double.TYPE)
        ).cast(Float.TYPE).returning()
```

or

```
.build(self, parameters) -> parameters.get(0).cast(Float.TYPE).returning()
```

dstepanov (changes requested):

dstepanov (approved):

PR #149, opened 2024-09-25
Creating toString, equals, and hashCode methods in Utils class

elifKurtay:
Feature implementation of [this issue](https://github.com/micronaut-projects/micronaut-sourcegen/issues/126#issue-2454329573).

Main changes:
- `@ToString` annotation for classes that returns the String description of the class through the following method:
`public static String [BeanName]Utils.toString([BeanName] object)`
- `@Equals` annotation for classes that returns the result of `equals` method through the following method:
`public static boolean [BeanName]Utils.equals([BeanName] object1, Object object2)`
- `@HashCode` annotation for classes that returns the hash value of the class instance through the following method:
`public static int [BeanName]Utils.hashCode([BeanName] object)`
- `@Secret` annotation for fields that the user wishes not to print in the toString function. Instead of printing the value, this annotation will print a fixed `"******"` representation. 

For `Person4` class in `test-suite-java` the generated Utils are:
```java
package io.micronaut.sourcegen.example;

import java.lang.Object;
import java.lang.String;
import java.lang.StringBuilder;
import java.util.Arrays;

public final class Person4Utils {
  public static String toString(Person4 object) {
    StringBuilder strBuilder = new java.lang.StringBuilder();
    strBuilder.append("Person4[");
    strBuilder.append("id=").append(object.getId()).append(", ");
    strBuilder.append("title=").append(object.getTitle()).append(", ");
    strBuilder.append("name=").append(object.getName()).append(", ");
    strBuilder.append("bytes=").append(java.util.Arrays.toString(object.getBytes())).append("]");
    return strBuilder.toString();
  }

  public static boolean equals(Person4 first, Object secondObject) {
    if (first == secondObject) {
      return true;
    }
    boolean isCorrectInstance = secondObject instanceof Person4;
    if (isCorrectInstance == false) {
      return false;
    }
    Person4 second = (Person4) (secondObject);
    boolean bothNullCondition = false;
    boolean equalsCondition = false;
    boolean isPropertyEqual = false;
    if (first.getId() != second.getId()) {
      return false;
    }
    bothNullCondition = first.getTitle() == null && second.getTitle() == null;
    equalsCondition = first.getTitle() != null && first.getTitle().equals(second.getTitle());
    isPropertyEqual = bothNullCondition || equalsCondition;
    if (isPropertyEqual == false) {
      return false;
    }
    bothNullCondition = first.getName() == null && second.getName() == null;
    equalsCondition = first.getName() != null && first.getName().equals(second.getName());
    isPropertyEqual = bothNullCondition || equalsCondition;
    if (isPropertyEqual == false) {
      return false;
    }
    bothNullCondition = first.getBytes() == null && second.getBytes() == null;
    equalsCondition = first.getBytes() != null && Arrays.equals(first.getBytes(), second.getBytes());
    isPropertyEqual = bothNullCondition || equalsCondition;
    if (isPropertyEqual == false) {
      return false;
    }
    return true;
  }

  public static int hashCode(Person4 object) {
    int hashValue = 1;
    hashValue = hashValue * 59 + (int) (object.getId() >>> 32 ^ object.getId());
    hashValue = hashValue * 59 + (int) (object.getTitle() == null ? 43 : object.getTitle().hashCode());
    hashValue = hashValue * 59 + (int) (object.getName() == null ? 43 : object.getName().hashCode());
    hashValue = hashValue * 59 + (int) (java.util.Arrays.hashCode(object.getBytes()));
    return hashValue;
  }
}
```


sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -538,8 +538,9 @@ private CodeBlock renderExpression(@Nullable ObjectDef objectDef, MethodDef meth
         }
         if (expressionDef instanceof ExpressionDef.Cast castExpressionDef) {
             return CodeBlock.concat(
-                CodeBlock.of("($T) ", asType(castExpressionDef.type(), objectDef)),
-                renderExpression(objectDef, methodDef, castExpressionDef.expressionDef())
+                CodeBlock.of("($T) (", asType(castExpressionDef.type(), objectDef)),
+                renderExpression(objectDef, methodDef, castExpressionDef.expressionDef()),
+                CodeBlock.of(")")
```
andriy-dmytruk, line 552: As Elif mentioned to me, it is hard to change order of operations in expressions and maybe a way of adding brackets would be good.

andriy-dmytruk (approved):

dstepanov:
Maybe instead of `MyBeanUtils` we can call it `MyBeanObject` or `MyBeanCompanion`.

```java
 public static String toString(Person4 object) {
    StringBuilder strBuilder = new java.lang.StringBuilder();
    strBuilder.append("Person4[");
    strBuilder.append("id=").append(object.getId()).append(", ");
    strBuilder.append("title=").append(object.getTitle()).append(", ");
    strBuilder.append("name=").append(object.getName()).append(", ");
    strBuilder.append("bytes=").append(java.util.Arrays.toString(object.getBytes())).append("]");
    return strBuilder.toString();
  }
```
You can add the first line in the constructor:
```java
    StringBuilder strBuilder = new java.lang.StringBuilder("Person4[");
```

```java
public static boolean equals(Person4 first, Object secondObject) {
    if (first == secondObject) {
      return true;
    }
    boolean isCorrectInstance = secondObject instanceof Person4;
    if (isCorrectInstance == false) {
      return false;
    }
    Person4 second = (Person4) (secondObject);
    boolean bothNullCondition = false;
    boolean equalsCondition = false;
    boolean isPropertyEqual = false;
    if (first.getId() != second.getId()) {
      return false;
    }
    bothNullCondition = first.getTitle() == null && second.getTitle() == null;
    equalsCondition = first.getTitle() != null && first.getTitle().equals(second.getTitle());
    isPropertyEqual = bothNullCondition || equalsCondition;
    if (isPropertyEqual == false) {
      return false;
    }
    bothNullCondition = first.getName() == null && second.getName() == null;
    equalsCondition = first.getName() != null && first.getName().equals(second.getName());
    isPropertyEqual = bothNullCondition || equalsCondition;
    if (isPropertyEqual == false) {
      return false;
    }
    bothNullCondition = first.getBytes() == null && second.getBytes() == null;
    equalsCondition = first.getBytes() != null && Arrays.equals(first.getBytes(), second.getBytes());
    isPropertyEqual = bothNullCondition || equalsCondition;
    if (isPropertyEqual == false) {
      return false;
    }
    return true;
  }
```

Please generate the code more or less like IntellJ equals generator, and avoid extra variables. Please also use `getClass` to compare classes. HashCode should have similar code like intellij:

```java
    @Override
    public boolean equals(Object o) {
        if (this == o) {
            return true;
        }
        if (o == null || getClass() != o.getClass()) {
            return false;
        }
        Person4 person4 = (Person4) o;
        return id == person4.id && title == person4.title && Objects.equals(name, person4.name) && Objects.deepEquals(bytes, person4.bytes);
    }

    @Override
    public int hashCode() {
        return Objects.hash(id, title, name, Arrays.hashCode(bytes));
    }
```

sourcegen-annotations/src/main/java/io/micronaut/sourcegen/annotations/Equals.java
```diff
@@ -0,0 +1,34 @@
...
+ * The method will be created in [BeanName]Utils class as a static method:
+ *      public static boolean BeanNameUtils.equals(BeanName this, Object other)
+ *
+ * @author Elif Kurtay
+ * @since 1.3
+ */
+
+@Retention(RUNTIME)
+@Target({ElementType.ANNOTATION_TYPE, ElementType.TYPE})
+public @interface Equals {
```
dstepanov: There is no point to have two annotation. Please have only one `EqualsAndHashCode`

sourcegen-annotations/src/main/java/io/micronaut/sourcegen/annotations/Secret.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * The annotation to be used with {@link ToString} on a property to hide the value from being printed.
+ *
+ * @author Elif Kurtay
+ * @since 1.3
+ */
+
+@Retention(RUNTIME)
+@Target({FIELD, PARAMETER})
+public @interface Secret {
```
dstepanov: I would prefer to have the same annotation as Lombok has (Exclude, Include) and have them as an inner annotation of `ToString` so you know where does it belong: `@ToString.Include`

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
dstepanov: You can probably check if the expression is an instance of a variable and exclude the brackets

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+    /*
+    Creates a toString method with signature:
+        public static String BeanNameUtils.toString(BeanName object)
+     */
+    private static void createToStringMethod(ClassDef.ClassDefBuilder classDefBuilder, String objectName, List<PropertyElement> properties) {
+        MethodDef method = MethodDef.builder("toString")
+            .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+            .returns(String.class)
+            .addParameter("object", ClassTypeDef.of(objectName))
+            .build((self, parameterDef) -> {
```
dstepanov: Rewrite this body as:

```java
ClassTypeDef.of(StringBuilder.class).instantiate().newLocal("strBuilder", variableDef -> {
      ExpressionDef exp = variableDef;
      for (PropertyElement propertyElement : properties) {
          exp = exp.invoke("append", variableDef.type(), ...);
      }
      return exp.invoke("toString", TypeDef.of(String.class)).returning();
});
```

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+                    statements.add(strBuilder.invoke(
+                        "append",
+                        ClassTypeDef.of(strBuilder.getClass()),
+                        ExpressionDef.constant(beanProperty.getSimpleName() + "=")
+                    ).invoke(
+                        "append",
+                        ClassTypeDef.of(strBuilder.getClass()),
+                        (TypeDef.of(beanProperty.getType()) instanceof TypeDef.Array) ?
+                            new ExpressionDef.CallStaticMethod(
+                                ClassTypeDef.of(Arrays.class),
```
dstepanov: Please fluent API:
`ClassTypeDef.of(Arrays.class).invokeStatic(...)`

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+    /*
+    Creates a hashCode method with signature:
+        public static int BeanNameUtils.hashCode(BeanName object)
+     */
+    private static void createHashCodeMethod(ClassDef.ClassDefBuilder classDefBuilder, String objectName, List<PropertyElement> properties) {
+        MethodDef method = MethodDef.builder("hashCode")
+            .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+            .addParameter("object", ClassTypeDef.of(objectName))
+            .returns(int.class)
+            .build((self, parameterDef) -> {
```
dstepanov: You should be able to write this as:

```java
  .build((self, parameterDef) -> {
      VariableDef objectVar = parameterDef.get(0).asVariable();
      return StatementDef.multi(
          objectVar.asCondition(" == ", self)
              .asConditionIf(ExpressionDef.trueValue().returning()),
          // TODO: if (o == null || getClass() != o.getClass()) {
          //            return false;
          //        }
          objectVar.cast(self.type()).newLocal("instance", variableDef -> {
              ExpressionDef exp = null;
              for (PropertyElement property : properties) {
                  ExpressionDef newEqualsExpression = ...;
                  if (exp == null) {
                      exp = newEqualsExpression;
                  } else {
                      exp = exp.asCondition(" && ", newEqualsExpression);
                  }
              }
              return exp.returning();
          })
  ));
```

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+                                new ExpressionDef.Condition(" ^ ", ExpressionDef.constant(32), thisProperty));
+                        } else if (typeName.equals("char")) {
+                            propertyHashExpression = new ExpressionDef.Condition(" - ", thisProperty, ExpressionDef.constant('0'));
+                        } else if (typeName.equals("short")) {
+                            propertyHashExpression = new ExpressionDef.Condition(" & ", thisProperty, ExpressionDef.constant(0xffff));
+                        } else { // for int and byte, return itself as an int
+                            propertyHashExpression = thisProperty;
+                        }
+                    } else { // OBJECT
+                        propertyHashExpression = new ExpressionDef.IfElse(
```
dstepanov: Please try to avoid using `new ExpressionDef` but use a a fluent API

dstepanov (changes requested):

dstepanov:
Please sign the License Agreement and add docs.

dstepanov:
We should consider to add as much as possible of what Lombok supports (can be added in another PR):

- https://projectlombok.org/api/lombok/EqualsAndHashCode
- https://projectlombok.org/api/lombok/ToString 

We should support `exclude`, `include`

The `callSuper` might be needed as well, in the case of false we should only include the properties in the current class

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+ *      {@link ToString} annotation for toString function
+ *      {@link Equals} annotation for equals function
+ *      {@link HashCode} annotation for hashCode function
+ *
+ * @author Elif Kurtay
+ * @since 1.3
+ */
+
+@Internal
+public final class UtilsAnnotationVisitor implements TypeElementVisitor<Object, Object> {
```
graemerocher: It feels like this implementation is doing too much. Ideally it would be split into separate visitors and add the annotation type to the generic arguments. So `EqualsAndHashCodeVisitor implements TypeElementVisitor<EqualsHashCode, Object>`, `ToStringVisitor implements TypeElementVisitor<ToString, Object>` etc. however I am aware that will require generating multiple Utils classes so maybe not ideal

Either way you need to override `getSupportedAnnotationNames()` and list the annotations this visitor processes.

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+            .addParameter("first", ClassTypeDef.of(objectName))
+            .addParameter("secondObject", TypeDef.of(Object.class))
+            .build((self, parameterDef) -> {
+                // local variables needed
+                VariableDef classname = new VariableDef.Local(objectName, ClassTypeDef.of(objectName));
+                List<StatementDef> statements = new ArrayList<>();
+                VariableDef firstObject = parameterDef.get(0).asVariable();
+                VariableDef.Local secondObject = new VariableDef.Local("second", classname.type());
+                VariableDef.Local bothNullCondition = new VariableDef.Local("bothNullCondition", TypeDef.of(boolean.class));
+                VariableDef.Local equalsCondition = new VariableDef.Local("equalsCondition", TypeDef.of(boolean.class));
```
graemerocher: use `var` on the left hand side of the statement to make these easier to read

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+                        continue;
+                    }
+
+                    // calculate new property hash value
+                    if (propertyTypeDef instanceof TypeDef.Array) {
+                        String methodName = (((TypeDef.Array) propertyTypeDef).dimensions() > 1) ?  "deepHashCode" : "hashCode";
+                        propertyHashExpression = arrays.invoke(methodName, TypeDef.of(int.class), thisProperty);
+                    } else if (propertyTypeDef instanceof TypeDef.Primitive) {
+                        String typeName = ((TypeDef.Primitive) propertyTypeDef).name();
+                        if (propertyTypeDef == TypeDef.BOOLEAN) {
```
graemerocher: rewrite these nested `ifs` as a switch expression that assigns `propertyHashExpression`

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+
+                // initialize hash value at 1
+                statements.add(new StatementDef.DefineAndAssign(hashValue, ExpressionDef.constant(1)));
+
+                // add all property values to hash calculations
+                TypeDef propertyTypeDef;
+                ExpressionDef thisProperty;
+                for (PropertyElement beanProperty : properties) {
+                    propertyTypeDef = TypeDef.of(beanProperty.getType());
+                    if (beanProperty.getReadMethod().isPresent()) {
```
graemerocher: extract `beanProperty.getReadMethod()` into a local variable and reuse

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+                    if (propertyTypeDef instanceof TypeDef.Primitive) {
+                        // ==, primitives do not need null check
+                        statements.add(new StatementDef.If(
+                            new ExpressionDef.Condition(" != ", firstProperty, secondProperty),
+                            ExpressionDef.constant(false).returning()
+                        ));
+                    } else {
+                        // .equals, check for double null or equal objects
+                        statements.add(new StatementDef.Assign(bothNullCondition,
+                                new ExpressionDef.Condition(" && ",
```
graemerocher: can we upstream improve the `ExpressionDef.Condition` type to make it easier to define `&&` expressions? //cc @dstepanov

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+                                    firstProperty.isNull(),
+                                    secondProperty.isNull())
+                            )
+                        );
+                        if (propertyTypeDef instanceof TypeDef.Array) {
+                            statements.add(new StatementDef.Assign(equalsCondition,
+                                    new ExpressionDef.Condition(" && ",
+                                        firstProperty.isNonNull(),
+                                        new ExpressionDef.CallStaticMethod(
```
graemerocher: Use `ExpressionDef.invokeStatic` instead

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+    private static void createHashCodeMethod(ClassDef.ClassDefBuilder classDefBuilder, String objectName, List<PropertyElement> properties) {
+        MethodDef method = MethodDef.builder("hashCode")
+            .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+            .addParameter("object", ClassTypeDef.of(objectName))
+            .returns(int.class)
+            .build((self, parameterDef) -> {
+                List<StatementDef> statements = new ArrayList<>();
+                ExpressionDef propertyHashExpression;
+                VariableDef.Local hashValue = new VariableDef.Local("hashValue", TypeDef.of(int.class));
+                VariableDef arrays = new VariableDef.Local("java.util.Arrays", ClassTypeDef.of(java.util.Arrays.class));
```
graemerocher: this seems an odd variable name, looks like a type name?

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
```diff
@@ -0,0 +1,406 @@
...
+                    } else if (propertyTypeDef instanceof TypeDef.Primitive) {
+                        String typeName = ((TypeDef.Primitive) propertyTypeDef).name();
+                        if (propertyTypeDef == TypeDef.BOOLEAN) {
+                            propertyHashExpression = new ExpressionDef.IfElse(
+                                thisProperty,
+                                ExpressionDef.constant(TRUE_HASH_VALUE),
+                                ExpressionDef.constant(FALSE_HASH_VALUE)
+                            );
+                        } else if (typeName.equals("float")) {
+                            VariableDef floatClass = new VariableDef.Local("Float", ClassTypeDef.of(float.class));
```
graemerocher: again non-consistent variable nam used. Also consider using constants for names

graemerocher (changes requested):
Great start! Well done @elifKurtay

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
dstepanov: All this should be written using the fluent API, for some basic conditions we can maybe add `and`/`or`

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
andriy-dmytruk: We should probably improve the docs by providing examples of how to do most common operations. Something like:

| Operation | Example code | Genereated example |
| --- | --- | --- |
| ... | ... | ... |
| Static invocation | `ClassTypeDef.of(String.class).invokeStatic("valueOf", ClassTypeDef.of(String.class), ExpressionDef.constant(2))` | `String.valueOf(2)` |

graemerocher: I think part of this PR is as we add more functionality we should extend the API of SourceGen to improve usability at the same time. It is an evolving API and clearly from the looks of this PR could do with enhancements to make writing code that writes code simpler.

dstepanov:
Please try to refactor it to look like my suggestions

andriy-dmytruk:
@dstepanov I don't think that `Objects.hash` is optimal. For primitives, it needs to wrap them to a class and then call `Object.hashCode`. Not that it is significant, but could be important, since it can be called frequently in collections. Lombok generates custom code for these types. The initial implementation was based on that.

Similarly for equals, using `==` seems like a very simple check, but could be slightly more optimal.

andriy-dmytruk:
> The callSuper might be needed as well, in the case of false we should only include the properties in the current class

We thought about this, but unfortunately we cannot call super from a different class. Generating methods using parent properties is possible, but not sure if that is desirable, as there could be custom behavior in the parent.

dstepanov:
I'm OK with inlining `Objects.equals` and `Objects.hash` which is:
```
    public static int hashCode(Object a[]) {
        if (a == null)
            return 0;

        int result = 1;

        for (Object element : a)
            result = 31 * result + (element == null ? 0 : element.hashCode());

        return result;
    }
```
I don't think we should copy exactly what Lobmok does without a reason.

> We thought about this, but unfortunately we cannot call super from a different class. Generating methods using parent properties is possible, but not sure if that is desirable, as there could be custom behavior in the parent.

That's not about calling the superclass; that's about including/excluding properties from the superclass.

elifKurtay:
> Please try to refactor it to look like my suggestions

I was still improving :') I will re-request review once I finished addressing all comments and requests 👍

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/UtilsAnnotationVisitor.java
elifKurtay: Should I try to do a basic implementation now or is this a thought for future? 

In my opinion, integrating parentheses into expressions and/or creating a `not` representation (`!(expression)`) would be more useful.

elifKurtay:
@andriy-dmytruk @dstepanov What is the verdict on `equals` and `hashCode` functions?

I think the current implementation of `equals` is good since it uses `==` for primitives and `Objects.equals` for objects. And apparently, `Objects.equals` check for both null before calling `.equals` on objects and `Array.equals` on arrays which is exactly equal to our initial implementation.

Should I go back to the initial description [here](https://github.com/micronaut-projects/micronaut-sourcegen/issues/126#issuecomment-2363808621) for `hashCode`? In addition, I think we can call the super class like:

```
public static int hashCode(Object object) {
        if (object == null)
            return 0;

        int result = object.getClass().getSuperclass().hashCode();
        
        result = 31 * result + (object.getElement() == null ? 0 : object.getElement().hashCode());
        ...

        return result;
    }
```

dstepanov:
I think you can inline `Objects.equals` which is `(a == b) || (a != null && a.equals(b))` and hashCode too

dstepanov:
`object.getClass().getSuperclass().hashCode()` this doesn't look correct

elifKurtay:
> We should consider to add as much as possible of what Lombok supports (can be added in another PR):
> 
> * https://projectlombok.org/api/lombok/EqualsAndHashCode
> * https://projectlombok.org/api/lombok/ToString
> 
> We should support `exclude`, `include`
> 
> The `callSuper` might be needed as well, in the case of false we should only include the properties in the current class

I am adding `Exclude` for both annotations but I think there are some things to consider when it comes to `Include`.
1. As we created a static method in a separate class, we are currently only using bean properties (a.k.a. properties with getters). We have no way of accessing methods or fields that are private and without a getter even if users put the `Include` annotation. So, I think we need to create a warning system for users otherwise it will seem like a bug. I can modify the implementation to check for public fields and methods but that's about it.
2. In Lombok, `Include` has two parameters: `name` and `rank`. Name is used to override the name displayed for the field/method. Rank is to modify processing order. This can be done for `ToString`, but it seems a bit useless in `EqualsAndHashCode` to me. Should I include these parameters or not?

For `callSuper`, there is again the only option of `object.getClass().getSuperclass()` which has a [`getDeclaredFields()`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredFields--) function. Though I never tried these methods, I can experiment. :)

Please let me know what do you think @dstepanov @graemerocher @andriy-dmytruk?

dstepanov:
> For callSuper, there is again the only option of object.getClass().getSuperclass() which has a [getDeclaredFields()](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredFields--) function. Though I never tried these methods, I can experiment. :)

That's something else, we would need to fetch properties that are only declared on the current bean. We can do it later.

elifKurtay:
> I think you can inline `Objects.equals` which is `(a == b) || (a != null && a.equals(b))` and hashCode too

This would require two/three variables since we don't have parentheses. Or I can "hack" it by casting the two sides to boolean for no reason 😄

dstepanov:
What do you mean by `we don't have parentheses`?

elifKurtay:
> What do you mean by `we don't have parentheses`?

Is there an ExpressionDef that adds parentheses around an expression like: `( <expression> )`?
asCondition does not add them as far as I tried. But maybe I am missing something?

dstepanov:
I see. You can add parentheses by default or introduce AND / OR conditions where you can add parentheses only for OR

sourcegen-annotations/src/main/java/io/micronaut/sourcegen/annotations/ToString.java
```diff
@@ -17,8 +17,7 @@
 
 import java.lang.annotation.*;
 
-import static java.lang.annotation.ElementType.FIELD;
-import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.*;
```
graemerocher: avoid star imports

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -54,10 +54,7 @@
 import java.io.IOException;
 import java.io.Writer;
 import java.lang.reflect.Array;
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
graemerocher: configure your IDE to avoid automatic star imports

elifKurtay:
Lastest changes:
- Exclude for ToString and EqualsAndHashCode (no Include for the time being.)
- And/Or Expressions along with asConditionAnd/ asConditionOr that adds parentheses to its expressions when the expression is a type of ExpressionDef.Condition/ExpressionDef.And/ExpressionDef.Or.
- Inline implementations of equals and hashCode

An example Person4.java object:
```
@ToString
@EqualsAndHashCode
public class Person4 {
    public enum Title {
        MRS,
        MR,
        MS
    }

    private long id;
    private Title title;
    private String name;
    private byte[] bytes;

    public Person4(long id, Title title, String name, byte[] bytes) {
        this.id = id;
        this.title = title;
        this.name = name;
        this.bytes = (bytes != null) ? Arrays.copyOf(bytes, bytes.length) : null;
    }

    public long getId() {
        return id;
    }

    public Title getTitle() {
        return title;
    }

    public String getName() {
        return name;
    }

    public byte[] getBytes() {
        return bytes;
    }

    @Override
    public String toString() {
        return Person4Object.toString(this);
    }

    @Override
    public boolean equals(Object o) {
        return Person4Object.equals(this, o);
    }

    @Override
    public int hashCode() {
        return Person4Object.hashCode(this);
    }
}
```

And its generated Person4Object class:
```
public final class Person4Object {
  public static String toString(Person4 instance) {
    StringBuilder strBuilder = new java.lang.StringBuilder("Person4[");
    return strBuilder.append("id=").append(instance.getId()).append(", ").append("title=").append(instance.getTitle()).append(", ").append("name=").append(instance.getName()).append(", ").append("bytes=").append(java.util.Arrays.toString(instance.getBytes())).append("]").toString();
  }

  public static boolean equals(Person4 instance, Object o) {
    if (instance == o) {
      return true;
    }
    if (o == null || instance.getClass() != o.getClass()) {
      return false;
    }
    Person4 other = (Person4) o;
    return (((instance.getId() == other.getId()) && ((instance.getTitle() == other.getTitle()) || ((instance.getTitle() != null) && instance.getTitle().equals(other.getTitle())))) && ((instance.getName() == other.getName()) || ((instance.getName() != null) && instance.getName().equals(other.getName())))) && ((instance.getBytes() == other.getBytes()) || ((instance.getBytes() != null) && Arrays.equals(instance.getBytes(), other.getBytes())));
  }

  public static int hashCode(Person4 instance) {
    if (instance == null) {
      return 0;
    }
    int hashValue = 1;
    hashValue = hashValue * 59 + (int) (instance.getId() >>> 32 ^ instance.getId());
    hashValue = hashValue * 59 + (int) (instance.getTitle() == null ? 43 : instance.getTitle().hashCode());
    hashValue = hashValue * 59 + (int) (instance.getName() == null ? 43 : instance.getName().hashCode());
    hashValue = hashValue * 59 + (int) (Arrays.hashCode(instance.getBytes()));
    return hashValue;
  }
}
```

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -581,6 +578,36 @@ private CodeBlock renderExpression(@Nullable ObjectDef objectDef, MethodDef meth
...
+                andExpressionDef.right() instanceof ExpressionDef.And || andExpressionDef.right() instanceof ExpressionDef.Or;
+            return CodeBlock.concat(
+                (isLeftCondition) ? CodeBlock.of("(") : CodeBlock.of(""),
+                renderExpression(objectDef, methodDef, andExpressionDef.left()),
+                (isLeftCondition) ? CodeBlock.of(")") : CodeBlock.of(""),
+                CodeBlock.of(" && "),
+                (isRightCondition) ? CodeBlock.of("(") : CodeBlock.of(""),
+                renderExpression(objectDef, methodDef, andExpressionDef.right()),
+                (isRightCondition) ? CodeBlock.of(")") : CodeBlock.of("")
+            );
```
graemerocher, line 592: split this into multiple `if/else` to make this easier to read and/or consider rewriting to use switch expressions that assign the code block

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -581,6 +578,36 @@ private CodeBlock renderExpression(@Nullable ObjectDef objectDef, MethodDef meth
...
+            boolean isRightCondition = orExpressionDef.right() instanceof ExpressionDef.Condition ||
+                orExpressionDef.right() instanceof ExpressionDef.And || orExpressionDef.right() instanceof ExpressionDef.Or;
+            return CodeBlock.concat(
+                (isLeftCondition) ? CodeBlock.of("(") : CodeBlock.of(""),
+                renderExpression(objectDef, methodDef, orExpressionDef.left()),
+                (isLeftCondition) ? CodeBlock.of(")") : CodeBlock.of(""),
+                CodeBlock.of(" || "),
+                (isRightCondition) ? CodeBlock.of("(") : CodeBlock.of(""),
+                renderExpression(objectDef, methodDef, orExpressionDef.right()),
+                (isRightCondition) ? CodeBlock.of(")") : CodeBlock.of("")
```
graemerocher: split this into multiple `if/else` to make this easier to read and/or consider rewriting to use switch expressions that assign the code block

graemerocher (changes requested):

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
elifKurtay: Wouldn't that be too many repeated lines? I am cool to change just thought it was too long for such small changes. 😄  

```
if (leftCondition && rightCondition) {
return CodeBlock.concat(
                CodeBlock.of("("),
                renderExpression(objectDef, methodDef, andExpressionDef.left()),
                CodeBlock.of(")"),
                CodeBlock.of(" && "),
                CodeBlock.of("("),
                renderExpression(objectDef, methodDef, andExpressionDef.right()),
                CodeBlock.of(")")
            );
} else if ( rightCondition) {
return CodeBlock.concat(
                renderExpression(objectDef, methodDef, andExpressionDef.left()),
                CodeBlock.of(" && "),
                CodeBlock.of("("),
                renderExpression(objectDef, methodDef, andExpressionDef.right()),
                CodeBlock.of(")")
            );
} else if (leftCondition) {
return CodeBlock.concat(
                CodeBlock.of("("),
                renderExpression(objectDef, methodDef, andExpressionDef.left()),
                CodeBlock.of(")"),
                CodeBlock.of(" && "),
                renderExpression(objectDef, methodDef, andExpressionDef.right()),
            );
} else {
return CodeBlock.concat(
                renderExpression(objectDef, methodDef, andExpressionDef.left()),
                CodeBlock.of(" && "),
                renderExpression(objectDef, methodDef, andExpressionDef.right()),
            );
} 
```

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -583,15 +589,39 @@ private CodeBlock renderExpression(@Nullable ObjectDef objectDef, MethodDef meth
...
-            return CodeBlock.concat(
-                (isLeftCondition) ? CodeBlock.of("(") : CodeBlock.of(""),
-                renderExpression(objectDef, methodDef, andExpressionDef.left()),
-                (isLeftCondition) ? CodeBlock.of(")") : CodeBlock.of(""),
-                CodeBlock.of(" && "),
-                (isRightCondition) ? CodeBlock.of("(") : CodeBlock.of(""),
-                renderExpression(objectDef, methodDef, andExpressionDef.right()),
-                (isRightCondition) ? CodeBlock.of(")") : CodeBlock.of("")
-            );
+            if (isLeftCondition && isRightCondition) {
```
dstepanov: This looks too complicated, can you just have List and add codeblock that are needed then call concat

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -234,67 +234,75 @@ private static void createHashCodeMethod(ClassDef.ClassDefBuilder classDefBuilde
...
+     */
+    private static ExpressionDef getPropertyHashValue(TypeDef propertyTypeDef, ExpressionDef thisProperty) {
+        ExpressionDef propertyHashCalculation;
+        // calculate new property hash value
+        if (propertyTypeDef.isArray()) {
+            String methodName = (((TypeDef.Array) propertyTypeDef).dimensions() > 1) ?  "deepHashCode" : "hashCode";
+            propertyHashCalculation = ClassTypeDef.of(Arrays.class).invokeStatic(methodName, TypeDef.of(int.class), thisProperty);
+        } else if (propertyTypeDef.isPrimitive()) {
+            String typeName = ((TypeDef.Primitive) propertyTypeDef).name();
+            if (propertyTypeDef == TypeDef.BOOLEAN) {
```
dstepanov: Use equals methods

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -234,67 +234,75 @@ private static void createHashCodeMethod(ClassDef.ClassDefBuilder classDefBuilde
...
+                propertyHashCalculation = thisProperty.asConditionIfElse(
+                    ExpressionDef.constant(TRUE_HASH_VALUE),
+                    ExpressionDef.constant(FALSE_HASH_VALUE)
+                );
+            } else if (typeName.equals("float")) {
```
dstepanov: Please add this primitives as constants to `TypeDef.Primitive` and move the TypeDef.BOOLEAN there, so we have `TypeDef.Primitive.BOOLEAN`

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -234,67 +234,75 @@ private static void createHashCodeMethod(ClassDef.ClassDefBuilder classDefBuilde
...
+                propertyHashCalculation = ClassTypeDef.of(Float.class).invokeStatic("floatToIntBits", TypeDef.of(int.class), thisProperty);
```
dstepanov: `TypeDef.of(int.class)` should be `TypeDef.Primitive.INT`

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
```diff
@@ -481,6 +494,19 @@ record NewInstance(ClassTypeDef type,
...
+    /**
+     * The new instance expression.
+     *
+     * @param type   The type
+     * @param value The initial value
+     * @author Denis Stepanov
+     * @since 1.0
+     */
+    @Experimental
+    record PrimitiveInstance(TypeDef type,
```
dstepanov: Please move this to `TypeDef.Primitive`, so we call `TypeDef.Primitive.INT.instantiate(1)`

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/TypeDef.java
```diff
@@ -249,13 +249,37 @@ default boolean isNullable() {
         return false;
     }
 
+    /**
+     * @return Is primitive type
+     */
+    default boolean isPrimitive() {
+        return this instanceof TypeDef.Primitive;
```
dstepanov, line 256: Override this new two methods

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/TypeDef.java
```diff
@@ -249,13 +249,37 @@ default boolean isNullable() {
...
         return this;
     }
 
+    /**
+     * Instantiate this class.
+     *
+     * @param value The initial value
+     * @return The instantiate expression
+     */
+    default ExpressionDef initialize(ExpressionDef value) {
```
dstepanov: This doesn't make sense considering TypeDef can be `TypeDef.Wildcard`

dstepanov (changes requested):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
```diff
@@ -569,6 +614,40 @@ public TypeDef type() {
...
+     * @param left     The left expression
+     * @param right    The right expression
+     * @author Elif Kurtay
+     * @since 1.3
+     */
+    @Experimental
+    record And(ExpressionDef left, ExpressionDef right) implements ExpressionDef {
+        @Override
+        public TypeDef type()  {
+            return TypeDef.of(boolean.class);
```
dstepanov: Can be a constant

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
```diff
@@ -569,6 +614,40 @@ public TypeDef type() {
...
+     * @param left     The left expression
+     * @param right    The right expression
+     * @author Elif Kurtay
+     * @since 1.3
+     */
+    @Experimental
+    record Or(ExpressionDef left, ExpressionDef right) implements ExpressionDef {
+        @Override
+        public TypeDef type()  {
+            return TypeDef.of(boolean.class);
```
dstepanov: Can be a constant

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+
+            List<PropertyElement> properties = element.getBeanProperties();
+
+            // create the utils functions if they are annotated
+            if (element.hasStereotype(ToString.class)) {
+                context.warn("@ToString annotation will only print out bean properties.", element);
+                createToStringMethod(objectBuilder, element.getSimpleName(), properties);
+            }
+            if (element.hasStereotype(EqualsAndHashCode.class)) {
+                createEqualsMethod(objectBuilder, element.getSimpleName(), properties);
```
dstepanov: Please pass `ClassElement` and create a type from it

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+                        } else {
+                            continue;
+                        }
+
+                        exp = exp.invoke("append", variableDef.type(),
+                                ExpressionDef.constant(beanProperty.getName() + "="))
+                            .invoke("append", variableDef.type(),
+                                (TypeDef.of(beanProperty.getType()).isArray()) ?
+                                    ClassTypeDef.of(Arrays.class).invokeStatic("toString",
+                                        TypeDef.of(String.class), List.of(propertyValue)) : propertyValue)
```
dstepanov: Create a constant for STRING

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+    /*
+    Creates an equals method with signature:
+        public static boolean BeanNameObject.equals(BeanName object1, Object object2)
+     */
+    private static void createEqualsMethod(ClassDef.ClassDefBuilder classDefBuilder, String objectName, List<PropertyElement> properties) {
+        MethodDef method = MethodDef.builder("equals")
+            .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+            .returns(boolean.class)
+            .addParameter("instance", ClassTypeDef.of(objectName))
+            .addParameter("o", TypeDef.of(Object.class))
```
dstepanov: Use constant for Object

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+                })
+            );
+        classDefBuilder.addMethod(method);
+    }
+
+    /*
+    Creates an equals method with signature:
+        public static boolean BeanNameObject.equals(BeanName object1, Object object2)
+     */
+    private static void createEqualsMethod(ClassDef.ClassDefBuilder classDefBuilder, String objectName, List<PropertyElement> properties) {
```
dstepanov: Replace `objectName` with `ClassTypeDef selfType`

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+                        for (PropertyElement beanProperty : properties) {
+                            propertyTypeDef = TypeDef.of(beanProperty.getType());
+                            if ( !beanProperty.hasAnnotation(EqualsAndHashCode.Exclude.class) && beanProperty.getReadMethod().isPresent()) {
+                                firstProperty = instance.invoke(beanProperty.getReadMethod().get(), List.of());
+                                secondProperty = variableDef.invoke(beanProperty.getReadMethod().get(), List.of());
+                            } else {
+                                continue;
+                            }
+                            ExpressionDef newEqualsExpression = propertyTypeDef.isPrimitive() ?
+                                firstProperty.asCondition(" == ", secondProperty)
```
dstepanov: Duplicate code

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+                                : firstProperty.asCondition(" == ", secondProperty)
+                                .asConditionOr(firstProperty.isNonNull().asConditionAnd(
+                                    (propertyTypeDef.isArray()) ?
+                                        ClassTypeDef.of(Arrays.class).invokeStatic("equals", TypeDef.BOOLEAN, Arrays.asList(firstProperty, secondProperty))
```
dstepanov: Why is it a list of values and not just `"equals", TypeDef.BOOLEAN, firstProperty, secondProperty`?

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+                VariableDef o = parameterDef.get(1).asVariable();
+                return StatementDef.multi(
+                    instance.asCondition(" == ", o)
+                        .asConditionIf(ExpressionDef.trueValue().returning()),
+                    o.isNull().asCondition(" || ",
+                        instance.invoke("getClass", ClassTypeDef.of("Class"))
+                            .asCondition(" != ", o.invoke("getClass", ClassTypeDef.of("Class"))))
+                        .asConditionIf(ExpressionDef.falseValue().returning()),
+                    o.cast(ClassTypeDef.of(objectName)).newLocal("other", variableDef -> {
+                        ExpressionDef exp = null;
```
dstepanov, line 200: Avoid variable definitions on the top

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+     */
+    private static void createHashCodeMethod(ClassDef.ClassDefBuilder classDefBuilder, String objectName, List<PropertyElement> properties) {
+        MethodDef method = MethodDef.builder("hashCode")
+            .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+            .addParameter("instance", ClassTypeDef.of(objectName))
+            .returns(int.class)
+            .build((self, parameterDef) -> StatementDef.multi(
+                    parameterDef.get(0).asExpression().isNull().asConditionIf(ExpressionDef.constant(0).returning()),
+                    TypeDef.of(int.class).initialize(ExpressionDef.constant(1)).newLocal("hashValue", hashValue -> {
+                        List<StatementDef> hashUpdates = new ArrayList<>();
```
dstepanov, line 251: Avoid variable definitions on the top

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+                        TypeDef propertyTypeDef;
+                        ExpressionDef propertyGetter;
+                        ExpressionDef propertyHashCalculation;
+
+                        for (PropertyElement beanProperty : properties) {
+                            propertyTypeDef = TypeDef.of(beanProperty.getType());
+                            if (!beanProperty.hasAnnotation(EqualsAndHashCode.Exclude.class) && beanProperty.getReadMethod().isPresent()) {
```
dstepanov: You keep repeating this kind of code everywhere, just check the value and continue; next block initialize a variable and get the value

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
```diff
@@ -0,0 +1,311 @@
...
+        ExpressionDef propertyHashCalculation;
+        // calculate new property hash value
+        if (propertyTypeDef.isArray()) {
+            String methodName = (((TypeDef.Array) propertyTypeDef).dimensions() > 1) ?  "deepHashCode" : "hashCode";
+            propertyHashCalculation = ClassTypeDef.of(Arrays.class).invokeStatic(methodName, TypeDef.of(int.class), property);
+        } else if (propertyTypeDef.isPrimitive()) {
+            String typeName = ((TypeDef.Primitive) propertyTypeDef).name();
+            if (propertyTypeDef == TypeDef.BOOLEAN) {
+                propertyHashCalculation = property.asConditionIfElse(
+                    ExpressionDef.constant(TRUE_HASH_VALUE),
```
dstepanov, line 282: Use conditions

dstepanov (changes requested):

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
elifKurtay: Like my previous implementation?
```
if (expressionDef instanceof ExpressionDef.And andExpressionDef) {
            boolean isLeftCondition = andExpressionDef.left() instanceof ExpressionDef.Condition ||
                andExpressionDef.left() instanceof ExpressionDef.And || andExpressionDef.left() instanceof ExpressionDef.Or;
            boolean isRightCondition = andExpressionDef.right() instanceof ExpressionDef.Condition ||
                andExpressionDef.right() instanceof ExpressionDef.And || andExpressionDef.right() instanceof ExpressionDef.Or;
            return CodeBlock.concat(
                (isLeftCondition) ? CodeBlock.of("(") : CodeBlock.of(""),
                renderExpression(objectDef, methodDef, andExpressionDef.left()),
                (isLeftCondition) ? CodeBlock.of(")") : CodeBlock.of(""),
                CodeBlock.of(" && "),
                (isRightCondition) ? CodeBlock.of("(") : CodeBlock.of(""),
                renderExpression(objectDef, methodDef, andExpressionDef.right()),
                (isRightCondition) ? CodeBlock.of(")") : CodeBlock.of("")
            );
 }
```
I changed this according to this review: https://github.com/micronaut-projects/micronaut-sourcegen/pull/149#discussion_r1778742157

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
elifKurtay: But this increases readability. I thought that is also important?

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/ObjectAnnotationVisitor.java
elifKurtay: Because there is support for both and (ExpressionDef... parameters) gets cast into a List anyway. But I can change

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
dstepanov: Something like this:
```
        List<CodeBlock> codeBlocks = new ArrayList<>();
            if (isLeftCondition) {
                codeBlocks.add(CodeBlock.of("("));
                codeBlocks.add(renderExpression(objectDef, methodDef, andExpressionDef.left()));
                codeBlocks.add(CodeBlock.of(")"));
            } else {
                codeBlocks.add(renderExpression(objectDef, methodDef, andExpressionDef.left());
            }
```
you can also create a new method to reduce the duplicity of checking for left and right

dstepanov (approved):

PR #151, opened 2024-10-01
Improve build method API + small improvements

dstepanov:
For the build method, returning a list of `VariableDef.MethodParameter` makes more sense
+ Small improvement in API after previous PR

### DIFF
--- a/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/MethodDef.java
+++ b/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/MethodDef.java
@@ -186,8 +186,11 @@ public final class MethodDef extends AbstractElement {
             return build((self, parameterDefs) -> null);
         }
 
-        public MethodDef build(BiFunction<VariableDef.This, List<ParameterDef>, StatementDef> bodyBuilder) {
-            StatementDef statement = bodyBuilder.apply(new VariableDef.This(TypeDef.THIS), parameters);
+        public MethodDef build(BiFunction<VariableDef.This, List<VariableDef.MethodParameter>, StatementDef> bodyBuilder) {
+            List<VariableDef.MethodParameter> variables = parameters.stream()
+                .map(ParameterDef::asVariable)
+                .toList();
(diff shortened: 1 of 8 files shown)


sourcegen-model/src/main/java/io/micronaut/sourcegen/model/MethodDef.java
```diff
@@ -186,8 +186,11 @@ public MethodDef build() {
             return build((self, parameterDefs) -> null);
         }
 
-        public MethodDef build(BiFunction<VariableDef.This, List<ParameterDef>, StatementDef> bodyBuilder) {
-            StatementDef statement = bodyBuilder.apply(new VariableDef.This(TypeDef.THIS), parameters);
+        public MethodDef build(BiFunction<VariableDef.This, List<VariableDef.MethodParameter>, StatementDef> bodyBuilder) {
+            List<VariableDef.MethodParameter> variables = parameters.stream()
+                .map(ParameterDef::asVariable)
+                .toList();
+            StatementDef statement = bodyBuilder.apply(new VariableDef.This(TypeDef.THIS), variables);
```
andriy-dmytruk, line 193: Out of curiosity, why do we need `ParameterDef` when we have `VariableDef.MethodParameter`? Just by name, they seem to share the same purpose.

andriy-dmytruk (approved):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/MethodDef.java
dstepanov: Essentially we have two APIs - one to define classes, methods, etc and ParameterDef is part of it; the second one is an API to define the code

graemerocher (approved):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/MethodDef.java
andriy-dmytruk: Makes sense now

PR #154, opened 2024-10-04
Implement more Kotlin generator, abstract common parts

dstepanov:
I had to abstract things like getting a property value, calling hashCode/equals/== to properly support Kotlin's version

### DIFF
--- a/sourcegen-generator-kotlin/src/main/kotlin/io/micronaut/sourcegen/KotlinPoetSourceGenerator.kt
+++ b/sourcegen-generator-kotlin/src/main/kotlin/io/micronaut/sourcegen/KotlinPoetSourceGenerator.kt
@@ -35,6 +35,8 @@ import com.squareup.kotlinpoet.javapoet.KotlinPoetJavaPoetPreview
 import com.squareup.kotlinpoet.javapoet.toKClassName
 import com.squareup.kotlinpoet.javapoet.toKTypeName
 import io.micronaut.core.annotation.Internal
+import io.micronaut.core.annotation.Nullable
+import io.micronaut.core.reflect.ClassUtils
 import io.micronaut.inject.visitor.VisitorContext
 import io.micronaut.sourcegen.generator.SourceGenerator
 import io.micronaut.sourcegen.model.AnnotationDef
@@ -42,8 +44,7 @@ import io.micronaut.sourcegen.model.ClassDef
 import io.micronaut.sourcegen.model.ClassTypeDef
 import io.micronaut.sourcegen.model.EnumDef
 import io.micronaut.sourcegen.model.ExpressionDef
-import io.micronaut.sourcegen.model.ExpressionDef.CallInstanceMethod
-import io.micronaut.sourcegen.model.ExpressionDef.CallStaticMethod
+import io.micronaut.sourcegen.model.ExpressionDef.*
 import io.micronaut.sourcegen.model.FieldDef
 import io.micronaut.sourcegen.model.InterfaceDef
 import io.micronaut.sourcegen.model.MethodDef
@@ -53,10 +54,13 @@ import io.micronaut.sourcegen.model.PropertyDef
 import io.micronaut.sourcegen.model.RecordDef
 import io.micronaut.sourcegen.model.StatementDef
 import io.micronaut.sourcegen.model.StatementDef.Assign
+import io.micronaut.sourcegen.model.StatementDef.DefineAndAssign
 import io.micronaut.sourcegen.model.TypeDef
+import io.micronaut.sourcegen.model.TypeDef.Primitive.PrimitiveInstance
 import io.micronaut.sourcegen.model.VariableDef
 import java.io.IOException
 import java.io.Writer
+import java.lang.reflect.Array
 import java.util.function.Consumer
 import javax.lang.model.element.Modifier
 
@@ -398,7 +402,7 @@ class KotlinPoetSourceGenerator : SourceGenerator {
             )
         }
         if (initializer != null) {
-            if (initializer is ExpressionDef.Constant) {
+            if (initializer is Constant) {
                 propertyBuilder.initializer(
                     CodeBlock.of(
                         "%L", initializer.value
@@ -437,7 +441,12 @@ class KotlinPoetSourceGenerator : SourceGenerator {
             .build()
     }
 
-    private fun buildNullableProperty(field: FieldDef, modifiers: Set<Modifier>, docs: List<String>, objectDef: ObjectDef?): PropertySpec {
+    private fun buildNullableProperty(
+        field: FieldDef,
+        modifiers: Set<Modifier>,
+        docs: List<String>,
+        objectDef: ObjectDef?
+    ): PropertySpec {
         return buildNullableProperty(
             field.name,
             field.type,
@@ -456,7 +465,7 @@ class KotlinPoetSourceGenerator : SourceGenerator {
             FunSpec.builder(method.name).returns(asType(method.returnType, objectDef))
         }
         funBuilder = funBuilder
-            .addModifiers(asKModifiers(modifiers))
+            .addModifiers(asKModifiers(method, modifiers))
             .addParameters(
                 method.parameters.stream()
                     .map { param: ParameterDef ->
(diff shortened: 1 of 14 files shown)


sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
```diff
@@ -337,6 +338,16 @@ default CallInstanceMethod invoke(MethodElement methodElement, List<ExpressionDe
...
+    /**
+     * The get property value expression.
+     *
+     * @param propertyElement The property element
+     * @return The get property value expression
+     * @since 1.3
+     */
+    default GetPropertyValue getPropertyValue(PropertyElement propertyElement) {
+        return new GetPropertyValue(this, propertyElement);
+    }
```
andriy-dmytruk, line 392: It would be good if `VariableDef.Field` would resolve the same as this automatically (inside class it would be just property access, otherwise a getter). Having 2 choices seems like a complication.

If we have to keep this, I think `PropertyValue` or `PropertyValueGetter` would be more matching with the naming here.

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
```diff
@@ -677,12 +688,98 @@ record NewArrayOfSize(TypeDef.Array type, int size) implements ExpressionDef {
...
+
+    /**
+     * The get hashCode expression.
+     *
+     * @param instance The instance
+     * @author Denis Stepanov
+     * @since 1.3
+     */
+    @Experimental
+    record HashCode(ExpressionDef instance) implements ExpressionDef {
```
andriy-dmytruk: Can you add a functional API:
```java
EpxressionDef.HashCode hashCode() {
  return new HashCode(this);
}
```
I think we should mention that the functional API is preferred, as well:
```java
/**
     * The get hashCode expression. Where possible use {@link ExpressionDef#hashCode} method
     * to create a hash code for an expression.
     *
     * @param instance The instance
     * @author Denis Stepanov
     * @since 1.3
     */
```

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
```diff
@@ -677,12 +688,98 @@ record NewArrayOfSize(TypeDef.Array type, int size) implements ExpressionDef {
...
+     * @since 1.3
+     */
+    @Experimental
+    record EqualsStructurally(ExpressionDef instance, ExpressionDef other) implements ExpressionDef {
+
+        @Override
+        public TypeDef type() {
+            return TypeDef.Primitive.BOOLEAN;
+        }
+    }
```
andriy-dmytruk, line 808: Can you add a functional API:
```java
EpxressionDef.EqualsStructurally equalsStructurally(ExpressionDef other) {
  return new EqualsStructurally(this, other);
}
```

sourcegen-generator-kotlin/src/main/kotlin/io/micronaut/sourcegen/KotlinPoetSourceGenerator.kt
```diff
@@ -589,119 +606,497 @@ class KotlinPoetSourceGenerator : SourceGenerator {
...
+            if (expressionDef is HashCode) {
+                val instanceExp = renderExpressionCode(objectDef, methodDef, expressionDef.instance)
+                val type = expressionDef.instance.type()
+                if (type.isArray) {
+                    if (type is TypeDef.Array && type.dimensions > 1) {
+                        return instanceExp.toBuilder().add(".contentDeepHashCode()").build()
+                    }
+                    return instanceExp.toBuilder().add(".contentHashCode()").build()
+                }
+                return instanceExp.toBuilder().add(".hashCode()").build()
```
andriy-dmytruk, line 973: `Objects.hashCode` implementation has a null case:
```
   public static int hashCode(Object o) {
        return o != null ? o.hashCode() : 0;
    }
```
but in this case we don't have it.

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
```diff
@@ -677,12 +688,98 @@ record NewArrayOfSize(TypeDef.Array type, int size) implements ExpressionDef {
...
+
+    /**
+     * The get class expression.
+     *
+     * @param instance The instance
+     * @author Denis Stepanov
+     * @since 1.3
+     */
+    @Experimental
+    record GetClassValue(ExpressionDef instance) implements ExpressionDef {
```
andriy-dmytruk: I think `ClassGetter` clould be better.

andriy-dmytruk (changes requested):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
dstepanov: There is a TODO; we don't support an alternative to `invoke` for fields.
We should have `SetPropertyValue` in the future so that's why I kept the `Get`

sourcegen-generator-kotlin/src/main/kotlin/io/micronaut/sourcegen/KotlinPoetSourceGenerator.kt
dstepanov: What do you mean?

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
dstepanov: I don't like `Getter` name. It reminds me of properties

sourcegen-generator-kotlin/src/main/kotlin/io/micronaut/sourcegen/KotlinPoetSourceGenerator.kt
andriy-dmytruk: I am not very familiar with kotlin sourcegen hrere, but for Java we have:
```java
           return renderExpression(objectDef, methodDef, instance.isNull().asConditionIfElse(
                ExpressionDef.constant(0),
                instance.invoke("hashCode", TypeDef.Primitive.BOOLEAN))
            );
```
Here we just call `o.hashCode()`?

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
andriy-dmytruk: You are right, getter is strange. Can you name it `GetClass` then? I was confused what is a "class value" when I initially read the name.

sourcegen-generator-kotlin/src/main/kotlin/io/micronaut/sourcegen/KotlinPoetSourceGenerator.kt
dstepanov: Yes, those methods are added externally, including a null check on `this`.

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
dstepanov: Added  `invokeHashCode` because we already have `invoke` for methods

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
dstepanov: Added

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
dstepanov: Added invokeGetClass

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
andriy-dmytruk: Makes sense

andriy-dmytruk (approved):

PR #158, opened 2024-10-23
Support generating delegate classes

andriy-dmytruk:
### Feature description

The change allows generating classes implementing delegation/wrapper pattern.

An example of generated class would be:
```java
public abstract class WorkerDelegate implements Worker {
  Worker delegatee;

  WorkerDelegate(Worker delegatee) {
    this.delegatee = delegatee;
  }

  @Override
  public double tasksPerDay() {
    return this.delegatee.tasksPerDay();
  }

  @Override
  public String name() {
    return this.delegatee.name();
  }

  @Override
  public boolean canComplete(List<String> tasks) {
    return this.delegatee.canComplete(tasks);
  }

  @Override
  public List<String> competencies() {
    return this.delegatee.competencies();
  }
}
```

User would then extend it to customize only some of the behavior, like:
```java
public class OvertimeWorker extends WorkerDelegate {

    OvertimeWorker(Worker delegatee) {
        super(delegatee);
    }

    @Override
    public double tasksPerDay() {
        return super.tasksPerDay() * 1.2;
    }
}
```

For Kotlin:
```kotlin
public abstract class WorkerDelegate : Worker {
  public var delegatee: Worker

  public constructor(delegatee: Worker) {
    this. delegatee = delegatee
  }

  public override fun tasksPerDay(): Double {
    return this. delegatee.tasksPerDay()
  }

  public override fun name(): String {
    return this. delegatee.name()
  }

  public override fun canComplete(tasks: List<String>): Boolean {
    return this. delegatee.canComplete(tasks)
  }

  public override fun competencies(): List<String> {
    return this. delegatee.competencies()
  }
}
```

### Issues

We do not seem to support generic parameters for methods very well, so that is not supported. Class generic parameters are also not supported.

### DIFF
--- a/sourcegen-annotations/src/main/java/io/micronaut/sourcegen/annotations/Delegate.java
+++ b/sourcegen-annotations/src/main/java/io/micronaut/sourcegen/annotations/Delegate.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2017-2021 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.sourcegen.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * The delegate annotation on a bean should create a delegate base class.
+ *
+ * @author Andriy Dmytruk
+ * @since 1.3
+ */
+@Documented
+@Retention(RUNTIME)
+@Target({ElementType.ANNOTATION_TYPE, ElementType.TYPE})
+public @interface Delegate {
+
+    /**
+     * Define what the class that delegate should be created for.
+     *
+     * @return The class to create delegate for
+     */
+    Class<?> type() default Void.class;
+
+}
--- a/sourcegen-generator-java/src/test/java/io/micronaut/sourcegen/javapoet/write/ConstructorWriteTest.java
+++ b/sourcegen-generator-java/src/test/java/io/micronaut/sourcegen/javapoet/write/ConstructorWriteTest.java
@@ -1,0 +1,71 @@
+package io.micronaut.sourcegen.javapoet.write;
+
+import io.micronaut.sourcegen.model.ClassDef;
+import io.micronaut.sourcegen.model.ParameterDef;
+import io.micronaut.sourcegen.model.PropertyDef;
+import io.micronaut.sourcegen.model.TypeDef;
+import org.junit.Test;
+
+import javax.lang.model.element.Modifier;
+import java.io.IOException;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+public class ConstructorWriteTest extends AbstractWriteTest {
+
+    @Test
+    public void writeAllFieldsConstructor() throws IOException {
+        ClassDef classDef = ClassDef.builder("test.Test")
+                .addModifiers(Modifier.PUBLIC)
+                .addProperty(PropertyDef.builder("name").ofType(String.class).build())
+                .addProperty(PropertyDef.builder("age").ofType(Integer.class).build())
+                .addAllFieldsConstructor()
+                .build();
+        var result = writeClassAndGetConstructor(classDef);
+
+        var expected = """
+        Test(String name, Integer age) {
+          this.name = name;
+          this.age = age;
+        }
+        """;
+        assertEquals(expected.strip(), result.trim());
+    }
+
+    @Test
+    public void writeNoFieldsConstructor() throws IOException {
+        ClassDef classDef = ClassDef.builder("test.Test")
+            .addModifiers(Modifier.PUBLIC)
+            .addProperty(PropertyDef.builder("name").ofType(String.class).build())
+            .addProperty(PropertyDef.builder("age").ofType(Integer.class).build())
+            .addNoFieldsConstructor(Modifier.PUBLIC)
+            .build();
+        var result = writeClassAndGetConstructor(classDef);
+
+        var expected = """
+        public Test() {
+        }
+        """;
+        assertEquals(expected.strip(), result.trim());
+    }
+
+    @Test
+    public void writeConstructor() throws IOException {
+        ClassDef classDef = ClassDef.builder("test.Test")
+            .addModifiers(Modifier.PUBLIC)
+            .addProperty(PropertyDef.builder("name").ofType(String.class).build())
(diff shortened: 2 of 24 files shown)


graemerocher:
looks good but needs docs. @dstepanov we should fix the issues with generic types

dstepanov:
Class generics are supported but method generics are not yet supported

graemerocher (approved):

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/DelegateAnnotationVisitor.java
```diff
@@ -0,0 +1,251 @@
...
+                    ).toList()
+                );
+            }
+            delegate.addSuperinterface(typeDef)
+                .addField(FieldDef.builder(DELEGATEE_MEMBER).ofType(typeDef).build())
+                .addAllFieldsConstructor();
+
+            Set<MethodSignature> createdMethods = new HashSet<>();
+            addDelegateMethods(element, delegate, typeDef, delegateType, createdMethods);
+            for (ClassElement interfaceElement : getAllInterfaces(element)) {
```
dstepanov: What is the point of getting interfaces? `getMethods` should return all methods from the class

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/DelegateAnnotationVisitor.java
```diff
@@ -0,0 +1,251 @@
...
+            }
+            for (ParameterElement parameter: method.getParameters()) {
+                ParameterDef def = ParameterDef.builder(
+                    parameter.getName(), TypeDef.of(parameter.getGenericType())
+                ).build();
+                methodBuilder.addParameter(def);
+                parameters.add(def.asVariable());
+            }
+            ExpressionDef.CallInstanceMethod delegateCall = ExpressionDef.invoke(new VariableDef.Field(
+                new This(delegateType), DELEGATEE_MEMBER, typeDef
```
dstepanov: use build method to access `this`

sourcegen-annotations/src/main/java/io/micronaut/sourcegen/annotations/Delegate.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * The delegate annotation on a bean should create a delegate base class.
+ *
+ * @author Andriy Dmytruk
+ * @since 1.3
+ */
+@Documented
+@Retention(RUNTIME)
+@Target({ElementType.ANNOTATION_TYPE, ElementType.TYPE})
+public @interface Delegate {
```
dstepanov, line 34: Another good use-case would be to generate a delegate interface with one abstract method `D delegate()`

sourcegen-generator-java/src/test/java/io/micronaut/sourcegen/javapoet/write/ConstructorWriteTest.java
```diff
@@ -0,0 +1,71 @@
...
+        assertEquals(expected.strip(), result.trim());
+    }
+
+    @Test
+    public void writeConstructor() throws IOException {
+        ClassDef classDef = ClassDef.builder("test.Test")
+            .addModifiers(Modifier.PUBLIC)
+            .addProperty(PropertyDef.builder("name").ofType(String.class).build())
+            .addProperty(PropertyDef.builder("age").ofType(Integer.class).build())
+            .addConstructor(List.of(ParameterDef.of("name", TypeDef.of(String.class))), Modifier.PROTECTED)
```
dstepanov, line 59: I would just use `addMethod` + `MethodDef.constructor`

dstepanov: So we don't need a new method

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/DelegateAnnotationVisitor.java
```diff
@@ -0,0 +1,251 @@
...
+            String delegateClassName = element.getPackageName() + "." + simpleName;
+
+            ClassTypeDef typeDef = ClassTypeDef.of(element);
+            ClassTypeDef delegateType = ClassTypeDef.of(delegateClassName);
+            ClassDefBuilder delegate = ClassDef.builder(delegateClassName)
+                .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT);
+
+            if (!element.getTypeArguments().isEmpty()) {
+                for (Entry<String, ClassElement> type: element.getTypeArguments().entrySet()) {
+                    if (type.getValue() instanceof GenericPlaceholderElement placeholderElement) {
```
dstepanov: Maybe create a new method `TypeVariable.of` to convert  `GenericPlaceholderElement` directly

dstepanov (changes requested):

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/DelegateAnnotationVisitor.java
andriy-dmytruk: I am not sure what you mean. I can see no alternative method for creating `this`.

sourcegen-generator-java/src/test/java/io/micronaut/sourcegen/javapoet/write/ConstructorWriteTest.java
andriy-dmytruk: I have added the `allFieldsConstructor` and `noFieldsConstructor` methods, so I think this one makes sense. It also simplifies the interface, as you do not need to pass the type being constructed again.

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/DelegateAnnotationVisitor.java
andriy-dmytruk: That's a good idea!

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/DelegateAnnotationVisitor.java
dstepanov: build method has a lamda that provides an access to this and parameters

sourcegen-generator-java/src/test/java/io/micronaut/sourcegen/javapoet/write/ConstructorWriteTest.java
dstepanov: NoFields might be the default constructor name

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/DelegateAnnotationVisitor.java
andriy-dmytruk: The best I could find is
```java
ExpressionDef.CallInstanceMethod delegateCall = new This(delegateType)
                .field(DELEGATEE_MEMBER, typeDef)
                .invoke(method.getName(), returnType, parameters);
```

dstepanov: Use `.build((self, parameterDefs) ->`

dstepanov:
@andriy-dmytruk Added some improvements to your code

andriy-dmytruk:
> @andriy-dmytruk Added some improvements to your code

Thanks!

PR #162, opened 2024-10-24
Add a common abstract class for type definitions (classes, records, interfaces, enums)

andriy-dmytruk:
The type stores methods and super interfaces. It could additionally store inner classes.


sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ObjectDefBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+        methods.add(method);
+        return thisInstance;
+    }
+
+    public final ThisType addSuperinterface(TypeDef superinterface) {
+        superinterfaces.add(superinterface);
+        return thisInstance;
+    }
+
+    public final VariableDef.This getThis() {
```
dstepanov: I don't think this is needed

dstepanov (approved):

Issue #160, opened 2024-10-24
Creating inner types inside classes


elifKurtay:
### Feature description

It would be nice to be able to create inner types inside classes. The current PropertyDef or FieldDef seems limited to creating variables and does not allow adding records, enums, or inner classes inside a class.

An example generation goal could be the following record:
```
public record LlamaRecord(
    @NotNull @Min(0) int age,
    Status status
) {
  enum Status {
    Single,
    Taken
  }
}
```

PR #164, opened 2024-10-25
Creating inner types

elifKurtay:
The implementation of [the issue #160](https://github.com/micronaut-projects/micronaut-sourcegen/issues/160#issue-2611898495)

- Records, enums, classes and interfaces can be created inside records, classes and interfaces using `.addInnerType(ObjectDef)`. 
- No inner type for enums.
- Inner types inside an interface automatically and explicitly gets the `public` and `static` modifiers.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -6,7 +6,7 @@ pluginManagement {
 }
 
 plugins {
-    id("io.micronaut.build.shared.settings") version "7.2.1"
+    id("io.micronaut.build.shared.settings") version "7.2.3"
 }
 
 rootProject.name = "sourcegen-parent"


sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -265,13 +293,36 @@ private void writeRecord(Writer writer, RecordDef recordDef) throws IOException
...
+                innerBuilder = getInterfaceBuilder(innerInterfaceDef);
+            } else if (innerType instanceof EnumDef innerEnumDef) {
+                innerBuilder = getEnumBuilder(innerEnumDef);
+            } else if (innerType instanceof RecordDef innerRecordDef) {
+                innerBuilder = getRecordBuilder(innerRecordDef);
+            } else {
+                throw new IllegalStateException("Unknown object definition: " + innerType);
+            }
+            if (isInterface) {
+                innerBuilder.addModifiers(Modifier.PUBLIC, Modifier.STATIC);
```
dstepanov: I think an interface is always static, and maybe public too

dstepanov:
Please also add tests that will generate inner classes

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
andriy-dmytruk, line 322: `Modifier.PUBLIC` should not be added. There. could be a case when someone wants a private interface.

Is `Modifier.STATIC` required? By default interfaces are static.

test-suite-java/src/test/java/io/micronaut/sourcegen/example/InnerTypesTest.java
```diff
@@ -0,0 +1,254 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class InnerTypesTest {
+
+    private static void assertAndClean(File classBuilder, String expectedString) throws IOException {
```
andriy-dmytruk: This test probably belongs in `sourcegen-generator/src/main/test/io/micronaut/sourcegen/write`. See other tests there.

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
andriy-dmytruk: Strangely in the test, the interface does not have these modifiers. Maybe java poet requires that you specify static explicitly?

But we should not set public modifier, as users might want the interface to be private in some cases.

test-suite-java/src/test/java/io/micronaut/sourcegen/example/InnerTypesTest.java
```diff
@@ -0,0 +1,254 @@
...
+        assertAndClean(generateClass(classBuilder.build()), expectedString);
+    }
+
+    @Test
+    public void classInRecord() throws IOException {
+        String expectedString = "record InnerRecord() {\n" +
+            "  class Inner {\n" +
+            "  }\n" +
+            "}\n";
+        ClassDef.ClassDefBuilder innerClassBuilder = ClassDef.builder("Inner");
```
andriy-dmytruk: For verification, can you add another test with modifiers to class, e.g. static and private?

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
elifKurtay: I believe in a record it also needs to be static like interface but it does not need explicit writing. Because only the interface wanted me to explicitly write both public and static modifiers.

graemerocher:
please resolve the conflicts

andriy-dmytruk:
> please resolve the conflicts

I moved common logic for `RecordDef`, `ClassDef`, `InterfaceDef` and `EnumDef` to `ObjectDef` abstract class. Please move the inner types property to `ObjectDef`, as well, as that is common logic for them.

PR #169, opened 2024-10-28
Adding inner types to objectDef

elifKurtay:
The implementation of [the issue #160](https://github.com/micronaut-projects/micronaut-sourcegen/issues/160#issue-2611898495)

- Records, enums, classes and interfaces can be created inside themselves using `.addInnerType(ObjectDef)`. 
- Inner types inside an interface automatically and explicitly gets the `public` and `static` modifiers.


test-suite-kotlin/src/test/kotlin/io/micronaut/sourcegen/example/InnerTypesGenerationTest.kt
```diff
@@ -0,0 +1,454 @@
+package io.micronaut.sourcegen.example
```
andriy-dmytruk: This should be in `sourcegen-generator-kotlin/test`

test-suite-kotlin/build.gradle.kts
```diff
@@ -10,6 +10,9 @@ dependencies {
     ksp(projects.sourcegenGeneratorKotlin)
     ksp(projects.testSuiteCustomGenerators)
     implementation(mn.micronaut.inject.kotlin)
+    implementation(projects.sourcegenGenerator)
+    implementation(projects.sourcegenGeneratorKotlin)
+    implementation(projects.sourcegenModel)
```
andriy-dmytruk: This is not needed.

andriy-dmytruk (approved):

test-suite-custom-generators/src/main/java/io/micronaut/sourcegen/custom/visitor/innerTypes/GenerateInnerTypeInEnumVisitor.java
```diff
@@ -0,0 +1,178 @@
...
+            .addEnumConstant("A")
+            .addEnumConstant("B")
+            .addEnumConstant("C")
+            .addInnerType(innerEnum)
+            .addInnerType(innerRecord)
+            .addInnerType(innerClass)
+            .addInnerType(innerInterface)
+            .addMethod(MethodDef.builder("myName")
+                .addModifiers(Modifier.PUBLIC)
+                .addStatement(new StatementDef.Return(
```
dstepanov: Please correct tests not to use `new` Statement / Expression, you should be able to write everything without it using flow style. See `build((aThis, parameters) -> ` use cases. In this case, it should be `aThis.invoke("string"...).returning()`

dstepanov (changes requested):

test-suite-custom-generators/src/main/java/io/micronaut/sourcegen/custom/visitor/innerTypes/GenerateInnerTypeInEnumVisitor.java
```diff
@@ -0,0 +1,153 @@
...
+            .addEnumConstant("C")
+            .addInnerType(innerEnum)
+            .addInnerType(innerRecord)
+            .addInnerType(innerClass)
+            .addInnerType(innerInterface)
+            .addMethod(MethodDef.builder("myName")
+                .addModifiers(Modifier.PUBLIC)
+                .returns(String.class)
+                .build((aThis, parameters) ->
+                        aThis.invoke("toString", TypeDef.STRING, List.of()).returning()))
```
dstepanov, line 92: Why do you keep using `List.of`? There is an array variation, and you can omit the parameters.

dstepanov (approved):

Issue #165, opened 2024-10-25
Support properties in EnumDef


andriy-dmytruk:
### Feature description

This would allow generating more complex enum types, like:
```java
@Serdeable
public enum Status(
    ACTIVE("active")
    IN_PROGRESS("in-progress"),
    DELETED("deleted");

    private final String value;

    public Status(String value) {
        this.value = value;
    }

    @JsonCreator
    public Status statusOf(String value) {
        return switch (value) {
            case "active" -> ACTIVE;
            case "in-progress" -> IN_PROGRESS;
            case "deleted" -> DELETED;
        };
    }

    @JsonValue
    public String getValue() {
        return value;
    }
) {
}
```

andriy-dmytruk:
As @elifKurtay pointed out, apart from generating properties, we will also need to support consturctors and storing parameters for them, like in `ACTIVE("active")`.

PR #170, opened 2024-10-29
Improving enum capabilities

elifKurtay:
Implementation of [this issue](https://github.com/micronaut-projects/micronaut-sourcegen/issues/165#issue-2614122027).

- using Java/Kotlin enum constant definition.
- checking enum constant naming.
- adding fields, properties, constructors and methods to enums.
- a **private** all fields constructor **needs to be added** on enums with constant object values.

Generated enum class example:
```
package io.micronaut.sourcegen.example;

import java.lang.String;

public enum MyEnum2 {

  A(0),
  B(1),
  C(2);

  public int myValue;

  private MyEnum2(int myValue) {
    this.myValue = myValue;
  }

  public String myName() {
    return this.toString();
  }
}
```

From the following definition:
```
EnumDef enumDef = EnumDef.builder(enumClassName)
            .addModifiers(Modifier.PUBLIC)
            .addEnumConstant("A", ExpressionDef.constant(0))
            .addEnumConstant("B", ExpressionDef.constant(1))
            .addEnumConstant("C", ExpressionDef.constant(2))
            .addField(FieldDef.builder("myValue").ofType(TypeDef.Primitive.INT).addModifiers(Modifier.PUBLIC).build())
            .addMethod(MethodDef.builder("myName")
                .addModifiers(Modifier.PUBLIC)
                .returns(TypeDef.STRING)
                .build((aThis, parameters) ->
                    aThis.invoke("toString", TypeDef.STRING, List.of()).returning()))
            .addAllFieldsConstructor(Modifier.PRIVATE)
            .build();
```

### DIFF
--- a/sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
+++ b/sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.sourcegen;
 
+import com.github.javaparser.utils.Pair;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.naming.NameUtils;
@@ -65,6 +66,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.IntStream;
+
+import static io.micronaut.sourcegen.javapoet.TypeSpec.anonymousClassBuilder;
 
 /**
  * The Java source generator.
@@ -154,8 +157,12 @@ public sealed class JavaPoetSourceGenerator implements SourceGenerator permits G
             enumBuilder.addAnnotation(asAnnotationSpec(annotation));
         }
 
-        for (String enumConstant : enumDef.getEnumConstants()) {
-            enumBuilder.addEnumConstant(enumConstant);
+        for (Pair<String,Object> enumConstant : enumDef.getEnumConstants()) {
+            if (enumConstant.b != null) {
+                enumBuilder.addEnumConstant(enumConstant.a, anonymousClassBuilder(enumConstant.b.toString()).build());
+            } else {
+                enumBuilder.addEnumConstant(enumConstant.a);
+            }
         }
 
         for (MethodDef method : enumDef.getMethods()) {
--- a/sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
+++ b/sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
@@ -19,7 +19,6 @@ import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.naming.NameUtils;
 import io.micronaut.core.reflect.ClassUtils;
-import io.micronaut.inject.ast.FieldElement;
 import io.micronaut.inject.ast.MemberElement;
 import io.micronaut.inject.ast.MethodElement;
 import io.micronaut.inject.ast.PropertyElement;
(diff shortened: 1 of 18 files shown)


dstepanov:
That example doesn't make sense.
It should be:
```
     .addEnumConstant(constant("active"))
```

dstepanov:
For enums we should add the concept of fields, properties and constructors we already have (which we already have because of methods)

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -593,9 +651,9 @@ private CodeBlock renderExpression(@Nullable ObjectDef objectDef, MethodDef meth
             if (memberElement instanceof MethodElement methodElement) {
                 return renderExpression(objectDef, methodDef, getPropertyValue.instance().invoke(methodElement));
             }
-            if (memberElement instanceof FieldElement fieldElement) {
+            // if (memberElement instanceof FieldElement fieldElement) {
                 // TODO: support field
-            }
+            // }
```
andriy-dmytruk: This was not implemented?

sourcegen-generator-kotlin/src/main/kotlin/io/micronaut/sourcegen/KotlinPoetSourceGenerator.kt
```diff
@@ -354,8 +329,17 @@ class KotlinPoetSourceGenerator : SourceGenerator {
...
+                    TypeSpec.anonymousClassBuilder()
+                        .addInitializerBlock(renderExpressionCode(null, MethodDef.builder("").build(), exp))
+                        .build()
+                )
+            } else {
+                enumBuilder.addEnumConstant(name)
+            }
         }
 
         var companionBuilder: TypeSpec.Builder? = null
```
andriy-dmytruk, line 319: Kotlin generator needs to be able to write properties as well, probably.
It would be best to copy the Java tests to make sure that everything works the same.

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -154,8 +156,64 @@ private void writeEnum(Writer writer, EnumDef enumDef) throws IOException {
...
+            field.getJavadoc().forEach(fieldBuilder::addJavadoc);
+            for (AnnotationDef annotation : field.getAnnotations()) {
+                fieldBuilder.addAnnotation(
+                    asAnnotationSpec(annotation)
+                );
+            }
+            enumBuilder.addField(
+                fieldBuilder
+                    .build()
+            );
```
andriy-dmytruk: To avoid duplicaton of this with ClassDef, it would probably make sense to have separate `buildProperties` and `buildFields` methods that are called for both of them. `TypeSpec.Builder` seems to be the same.

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
elifKurtay: It came from master, so I did not want to touch it.

andriy-dmytruk (approved):
Great work, everything looks good to me!

sourcegen-generator-java/src/main/java/io/micronaut/sourcegen/JavaPoetSourceGenerator.java
```diff
@@ -168,9 +169,24 @@ private TypeSpec.Builder getEnumBuilder(EnumDef enumDef) {
...
+        enumDef.getEnumConstants().forEach((name, exps) -> {
+            if (exps != null) {
+                CodeBlock.Builder expBuilder = CodeBlock.builder();
+                for (int i = 0; i < exps.size(); i++) {
+                    expBuilder.add(renderExpression(null, null, exps.get(i)));
+                    if (i < exps.size() - 1) {
+                        expBuilder.add(", ");
+                    }
+                }
+                enumBuilder.addEnumConstant(name, anonymousClassBuilder(expBuilder.build()).build());
```
dstepanov, line 181: It's a bit strange that a enum is defined this way in the javapoet enum

dstepanov (approved):

PR #196, opened 2024-11-28
Support inner types and enums with custom constructors for bytecode generator

dstepanov:


test-suite-bytecode/src/test/java/io/micronaut/sourcegen/example/InnerTypesTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+//        MyEnumWithInnerTypes.InnerRecord innerRecord = MyEnumWithInnerTypes$InnerRecordBuilder.builder().id(3).build();
+//        assertEquals(3, innerRecord.id());
+
+        MyEnumWithInnerTypes.InnerClass innerClass = new MyEnumWithInnerTypes.InnerClass("name");
+        assertEquals("name", innerClass.getName());
+    }
+
+//    @Test
+//    public void recordTest() {
```
dstepanov, line 46: Records not supported yet, keeping the commented out tests for the future

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ClassTypeDef.java
```diff
@@ -54,14 +54,24 @@ public sealed interface ClassTypeDef extends TypeDef {
      * @return The simple name
      */
     default String getSimpleName() {
-        return NameUtils.getSimpleName(getName());
+        String name = getCanonicalName();
+        int i = name.lastIndexOf('.');
+        if (i == -1) {
+            return name;
+        }
+        return name.substring(i + 1);
```
elifKurtay, line 62: return name.substring(name.lastIndexOf('.') + 1);

elifKurtay:
Right now when we annotate any TypeDef, we cannot reach the Type's information inside AnnotatedClassTypeDef/AnnotatedTypeDef. It might be nice to add getters to reach the typeDef inside or copy the ClassTypeDef's default functions also here like in Parameterized.

elifKurtay (approved):

graemerocher (approved):

dstepanov:
> Right now when we annotate any TypeDef, we cannot reach the Type's information inside AnnotatedClassTypeDef/AnnotatedTypeDef. It might be nice to add getters to reach the typeDef inside or copy the ClassTypeDef's default functions also here like in Parameterized.

I'm not sure what you mean. In most cases, I only differentiate whether the type is `TypeDef` or `ClassTypeDef`; that's why we have two annotated variations. What is precisely the use case?

PR #200, opened 2024-12-03
Generator improvements

dstepanov:
- Refactoring of conditions, `isXyz` syntax returns a conditional expression `ifXyz` allows to direct create an if/else statement/expression - make the syntax much better
- Moved main conditional methods to `ConditionExpressionDef` requiring `or`/`and` to use a condition
- Now it's required to convert blind expression to a condition by calling `isTrue`/`isFalse`/`IsNull` etc
- Fixes for bytecode generator + support for adding correct variables and parameters names for better reading decompiled classes
- Bytecode checker turn off by default

### DIFF
--- a/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
+++ b/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
@@ -85,7 +85,7 @@ public sealed interface ExpressionDef
      * @return The condition expression
      * @since 1.2
      */
-    default ExpressionDef asCondition(String op, ExpressionDef expression) {
+    default ConditionExpressionDef asCondition(String op, ExpressionDef expression) {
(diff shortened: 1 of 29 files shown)


graemerocher (approved):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
```diff
@@ -205,24 +183,94 @@ default StatementDef isNull(StatementDef ifStatement) {
...
+    /**
+     * Is true - if / else statement.
+     *
+     * @param ifStatement   If statement
+     * @param elseStatement Else statement
+     * @return Is true statement
+     * @since 1.5
+     */
+    default StatementDef ifTrue(StatementDef ifStatement, StatementDef elseStatement) {
+        return isTrue().doIfElse(ifStatement, elseStatement);
```
andriy-dmytruk, line 230: Nice, this is great!

PR #215, opened 2024-12-28
Fixes gradle scripts

altro3:
Gradle 8.12
Update libs

Fixed tests for windows


sdelamo:
@graemerocher @altro3 I think this broke the published artifacts https://github.com/micronaut-projects/micronaut-sourcegen/issues/221

sdelamo:
moreover, it would be good not to change the project structure. We use `buildSrc` in every repo. This PR changes to use `build-logic`. We have so many repositories that if the structure is different in every repo it is a nightmare to maintain.

build-logic/src/main/groovy/io.micronaut.build.internal.sourcegen-module.gradle
```diff
@@ -0,0 +1,10 @@
+plugins {
+    id("io.micronaut.build.internal.sourcegen-base")
+    id("io.micronaut.build.internal.module")
+}
+
+micronautBuild {
+    binaryCompatibility {
+        enabled.set(false)
```
sdelamo, line 8: why are we disabling binary compatibility?

altro3: @sdelamo don't know, that's not my code. I didn't change this file, just move it to build-logic

dstepanov: This needs to be always disable for a new project

PR #226, opened 2025-01-13
Support all math and compare operations

dstepanov:


graemerocher (approved):
Might be nicer as `.math().plus(otherExpression)` but probably fine like this too

dstepanov:
I was thinking to do `.number().plus(otherExpression)` maybe will add it in the future

Issue #209, opened 2024-12-12
Generate Maven and Gradle plugin code stubs


andriy-dmytruk:
# Description 

There is a lot of maven and gradle-specific code for micronaut plugins that seems to be generateable. An example is micronaut openapi generators. We will also need plugins for JSON schema generations and possibly others in the future.

For openapi all the parameters are repeated 3 times currently:
- In the entrypoint to the openapi generator: https://github.com/micronaut-projects/micronaut-openapi/blob/6.13.x/openapi-generator/src/main/java/io/micronaut/openapi/generator/MicronautCodeGeneratorEntryPoint.java (there are separate options, option builders and their implementation to simplify the usage in plugins).
- In the maven plugin: https://github.com/micronaut-projects/micronaut-maven-plugin/blob/4.7.x/micronaut-maven-plugin/src/main/java/io/micronaut/maven/openapi/AbstractOpenApiMojo.java. The parameters are fed into the entrypoint and generation is called.
- In the gradle plugin task: https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/AbstractOpenApiGenerator.java. But also in the gradle plugin work action and extension.

Generating would have the advanatage that both versions would have consistent parameters and possibly similar API.

# Generation description

Add a new module to generate plugins.

Given a configuration for an plugin executable, like
```java
/**
 * Test configuration.
 *
 * @param packageName The package name to generate to
 */
@PluginConfig(mavenPropertyPrefix = "test")
public record TestConfig(
   @PluginParameter(mavenProperty = "packageName", required = true)
   String packageName,
   @PluginParameter(defaultValue = false)
   boolean serverGeneration,
   Map<String, String> additionalProperties,
   @PluginParameter(internal = true)
   String outputDirectory
) implements PluginExecutable {

   public void execute() {
       doWork(packageName, serverGeneration, additionalProperties, outputDirectory);
   }

}
```

we would want to generate plugin code stubs for reading these properties and calling the execute method.

### Maven
we could generate maven Mojo:
```java
public abstract class AbstractTestMojo extends AbstractMicronautMojo {

    /**
     * The package name to generate to.
     */
    @Parameter(property = "test.packageName", required = true)
    protected String packageName;

    @Parameter(defaultValue = "false")
    protected boolean serverGeneration;
    
    @Parameter
    protected Map<String, Object> additionalProperties;
    
[...]

PR #232, opened 2025-01-17
Prototype plugin stubs generation #209

andriy-dmytruk:
This adds support for generating Maven and Gradle stubs (#209).

## Description

A plugin task configuration can be provided in a micronaut repo, like `micronaut-json-schema`. It is marked and configured using `@PluginTask` annotation. Then, `@GenerateGradlePlugin` or `@GenerateMavenMojo` annotations can be used in plugin repositories to generate plugin sources.

This has the advantage:
* Plugin parameters can be defined in a single place and do not need to be repeated in plugins (currently we repeat parameters once for Maven and three times for Gradle).
* This makes sure that parameter names and parameter defaults are consistent between plugins reducing manual errors. 
* The documentation needs to be written only once improving consistency again.

As an example, OpenAPI plugin has more than 40 parameters, which can be added and modified. This would simplify that effort.

## Example

Example is provided in the `test-suite-plugin-`.
A plugin task can be configured with
```java
@PluginTask
public record GenerateSimpleRecordTask(
    @PluginTaskParameter(required = true, globalProperty = "typeName")
    String typeName,
    @PluginTaskParameter(defaultValue = "1", globalProperty = "version")
    Integer version,
    @PluginTaskParameter(defaultValue = "com.example", globalProperty = "packageName")
    String packageName,
    Map<String, String> properties,
    List<String> javadoc,
    @PluginTaskParameter(output = true, directory = true, required = true)
    File outputFolder
) {

    /**
     * Generate a simple record in the supplied package and with the specified version.
     * This javadoc will be copied to the respected plugin implementations.
     */
    @PluginTaskExecutable
    public void generateSimpleRecord() {
       /* ... */
    }
}
```

Then gradle and maven sources can be generated by triggering annotations:
```java
@GenerateGradlePlugin(
    namePrefix = "Test",
    micronautPlugin = false,
    types = {
        Type.GRADLE_TASK,
        Type.GRADLE_EXTENSION,
        Type.GRADLE_SPECIFICATION,
        Type.GRADLE_PLUGIN
    },
    tasks = {
        @GenerateGradleTask(
            namePrefix = "GenerateSimpleRecord",
            extensionMethodName = "generateSimpleRecord",
            source = "io.micronaut.sourcegen.example.plugin.GenerateSimpleRecordTask"
        ),
        @GenerateGradleTask(
            namePrefix = "GenerateSimpleResource",
            extensionMethodName = "generateSimpleResource",
            source = "io.micronaut.sourcegen.example.plugin.GenerateSimpleResourceTask"
        )
    }
)
public final class GeneratePluginTrigger {
}
```
```java
@GenerateMavenMojo(
    namePrefix = "AbstractGenerateSimpleRecord",
    micronautPlugin = false,
    source = "io.micronaut.sourcegen.example.plugin.GenerateSimpleRecordTask",
    mavenPropertyPrefix = "test.generate.simple.record"
)
@GenerateMavenMojo(
    namePrefix = "AbstractGenerateSimpleResource",
    micronautPlugin = false,
    source = "io.micronaut.sourcegen.example.plugin.GenerateSimpleResourceTask",
    mavenPropertyPrefix = "test.generate.simple.resource"
)
public final class GenerateMojoTrigger {
}
```
This generates Gradle plugin, extension, specifications and tasks. For maven it generates maven mojos.
Test suite provides a more complete example. 

Gradle extension and maven mojos are extended with implementation-specific details and also tests are shown.


sourcegen-plugin-annotations/build.gradle.kts
```diff
@@ -0,0 +1,9 @@
+plugins {
+    id("io.micronaut.build.internal.sourcegen-module")
+}
+
+micronautBuild {
+    binaryCompatibility {
+        enabled.set(false)
```
melix: :x: Should use `enabledAfter("some version")` instead

test-suite-plugin-gradle/src/test/java/io/micronaut/sourcegen/example/plugin/gradle/TestPluginTest.java
```diff
@@ -0,0 +1,129 @@
...
+            generateRecordWithName("MyRecord", "io.micronaut.test", spec -> {
+                spec.getJavadoc().add("A simple record")
+                spec.getProperties().put("title", "java.lang.String")
+                spec.getProperties().put("age", "java.lang.Integer")
+            })
+            generateResource("generateHello", "META-INF/hello.txt", "Hello!");
+        }
+
+        repositories {
+            mavenLocal()
```
melix: Should probably avoid this.

sourcegen-plugin-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/gradle/builder/GradleTaskBuilder.java
```diff
@@ -0,0 +1,316 @@
...
+                propBuilder.addAnnotation("org.gradle.api.tasks.Input");
+                if (parameter.source().getType().isAssignable(File.class)) {
+                    if (parameter.directory()) {
+                        propBuilder.addAnnotation(AnnotationDef.builder(ClassTypeDef.of("org.gradle.api.tasks.InputDirectory")).build());
+                    } else {
+                        propBuilder.addAnnotation(AnnotationDef.builder(ClassTypeDef.of("org.gradle.api.tasks.InputFile")).build());
+                    }
+                    propBuilder.addAnnotation(AnnotationDef.builder(ClassTypeDef.of("org.gradle.api.tasks.PathSensitive"))
+                        .addMember("value", ClassTypeDef.of("org.gradle.api.tasks.PathSensitivity")
+                            .getStaticField("NONE", TypeDef.of("org.gradle.api.tasks.PathSensitivity"))
```
melix: :x: This should probably be configurable. No path sensitivity means that it doesn't matter if the file name changes, or the path to the files change, which is unlikely.

sourcegen-plugin-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/gradle/builder/GradleTaskBuilder.java
```diff
@@ -0,0 +1,316 @@
...
+        }
+        return objects;
+    }
+
+    private List<ObjectDef> buildTask(String packageName, GradleTaskConfig taskConfig) {
+        String taskType = packageName + "." + taskConfig.namePrefix() + TASK_SUFFIX;
+        ClassDefBuilder builder = ClassDef.builder(taskType)
+            .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
+            .superclass(ClassTypeDef.of("org.gradle.api.DefaultTask"))
+            .addAnnotation("org.gradle.api.tasks.CacheableTask")
```
melix: We should probably make this configurable. It may not be worth caching, or, caching may be broken for whatever reason.

melix (commented):
That's fairly impressive work already! I haven't performed a full review yet, only tried to understand the basics. What would probably help is some unit tests which assert the generated sources: currently the test suite projects use the generated code, but we can't "see" it so it's hard to tell how "correct" the generated code is. For example, this generates plugin code, but I'd like to understand better how well it integrates.

sourcegen-plugin-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/gradle/builder/GradleTaskBuilder.java
andriy-dmytruk: You are right, I copied from some unusual property I saw

andriy-dmytruk:
@melix Yes, you are right. I added some simple tests that verify the generated sources

sourcegen-plugin-annotations/src/main/java/io/micronaut/sourcegen/annotations/GenerateGradlePlugin.java
```diff
@@ -106,6 +109,17 @@ enum Type {
...
+        /**
+         * @return Whether the task is cacheable
+         */
+        boolean cacheable() default true;
+
+        /**
+         * @return Path sensitivity to use for file parameters must be one of
+         * {@code org.gradle.api.tasks.PathSensitivity} values: ABSOLUTE, RELATIVE, NAME_ONLY, NONE.
+         */
+        String pathSensitivity() default "ABSOLUTE";
```
graemerocher: should this be an enum?

settings.gradle.kts
```diff
@@ -21,13 +21,18 @@ include("sourcegen-generator-kotlin")
 include("sourcegen-generator-bytecode")
 include("sourcegen-bytecode-writer")
 include("sourcegen-bom")
+include("sourcegen-plugin-annotations")
+include("sourcegen-plugin-generator")
```
graemerocher: possibly rename prefix to `sourcegen-build-plugin-`

sourcegen-plugin-annotations/src/main/java/io/micronaut/sourcegen/annotations/GenerateGradlePlugin.java
melix: This should actually be per property, not a single entry for all paths.

graemerocher:
can you address the Sonar failures?

dstepanov:
Could you move this to a different project? I don't think we should put everything related to the source generation to this project.

graemerocher:
good point, might make sense in a new project.

andriy-dmytruk:
Closing in favor of https://github.com/micronaut-projects/micronaut-build-plugin-sourcegen/pull/2

PR #269, opened 2025-03-18
Add support for lambda expressions

andriy-dmytruk:
Add support for Java lambda expressions


sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/expression/LambdaExpressionWriter.java
```diff
@@ -0,0 +1,235 @@
...
+
+    @Override
+    public void write(GeneratorAdapter generatorAdapter, MethodContext context) {
+        List<VariableDef> capturedVariables = captureVariables(lambda.method());
+        MethodDef implementationMethodDef = createLambdaMethodDef(context, lambda, capturedVariables);
+        context.lambdaMethods().add(implementationMethodDef);
+
+        // The captured variables are the parameters to the called bootstrap method
+        for (VariableDef variable: capturedVariables) {
+            System.out.println("Writing variable: " + variable);
```
graemerocher: remove this println

sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/expression/VariableExpressionWriter.java
```diff
@@ -32,6 +32,13 @@ public VariableExpressionWriter(VariableDef variableDef) {
 
     @Override
     public void write(GeneratorAdapter generatorAdapter, MethodContext context) {
+        System.out.println("Variable: " + variableDef);
+        System.out.println("isLambda: " + context.isLambda());
```
graemerocher: remove printlns

sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/expression/VariableExpressionWriter.java
```diff
@@ -82,4 +77,56 @@ public void write(GeneratorAdapter generatorAdapter, MethodContext context) {
...
+            name = local.name();
+        } else if (variableDef instanceof VariableDef.MethodParameter methodParameter) {
+            name = methodParameter.name();
+        }
+
+        if (name != null) {
+            writeParameterVariable(generatorAdapter, context, name);
+            return;
+        }
+        throw new UnsupportedOperationException("Unrecognized lambda variable: " + variableDef);
```
graemerocher, line 126: the processingexception instead

graemerocher (changes requested):
nice! Could you add support for lambda references too? Like `Object::toString`

sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/expression/VariableExpressionWriter.java
andriy-dmytruk: There isn't always a visitor context available (the classes are not always generated during annotation processing).

andriy-dmytruk:
> nice! Could you add support for lambda references too? Like `Object::toString`

Yes, that would be good, too. Seems like a lot of changes because there are 3 different types of it: 
* `Object::toString` invokes `param.toString()`
* `MyType::convertToString` will invoke `MyType.convertToString(param)` if `convertToString` is static.
* `this::convertToString` will invoke `this.convertToString(param)`. There is also `var::convertToString` which is similar.

So probably best in a different PR.

But it seems like internally all of them are implemented by just creating an appropriate lambda, e.g. this method just calls param.toString():
```java
 private static java.lang.String lambda$callLambda$0(java.lang.String);
    descriptor: (Ljava/lang/String;)Ljava/lang/String;
    flags: (0x100a) ACC_PRIVATE, ACC_STATIC, ACC_SYNTHETIC
    Code:
      stack=2, locals=1, args_size=1
         0: aload_0
         1: iconst_1
         2: invokevirtual #50                 // Method java/lang/String.substring:(I)Ljava/lang/String;
         5: areturn

```

graemerocher (approved):

dstepanov:
I will check

dstepanov:
Now this code is a bit strange:
```
      public static Lambda of(ClassTypeDef lambdaType, MethodDef target, MethodBodyBuilder bodyBuilder) {
            return new Lambda(
                lambdaType,
                target,
                MethodDef.builder(target.getName())
                    .returns(target.getReturnType())
                    .addParameters(target.getParameters())
                    .build(bodyBuilder)
            );
        }
```
Considering that we implement lamda and in a case of `Function` the method is generic and the target method should be with raw types and the other with resolved generics. So we need to think about it a bit, maybe we can deduct the types based on the statements which we already support a bit, or have a way to specify generics of the method.

andriy-dmytruk:
> Now this code is a bit strange

Yes, seems like it doesn't cover every case. Deducing return type would probably work, but we can't deduce parameters. 

I also don't really like that we expect `MethodDef target`, as user might pass incorrect value, which will work for the java generator, but won't work for bytecode. Deducing this from ObjectDef or ClassElement as done in other methods is less error-prone.

andriy-dmytruk:
@dstepanov Is this in a good state now?

dstepanov (approved):

PR #272, opened 2025-03-19
Add support for Lambda expressions in Kotlin

andriy-dmytruk:
- Modify kotlin generator
- Add StringConcatenation ExpressionDef type
- Add support for functional interface through annotation

It has some additional changes, so I created a separate PR

### DIFF
--- a/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
+++ b/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
@@ -47,8 +47,7 @@ import java.util.stream.Stream;
  * @since 1.0
  */
 @Experimental
-public sealed interface ExpressionDef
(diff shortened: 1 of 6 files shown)


sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
```diff
@@ -1205,6 +1222,28 @@ public enum OpType {
...
+    /**
+     * The string concatenation operation.
+     *
+     * @param left     The left expression
+     * @param right    The right expression
+     * @author Andriy Dmytruk
+     * @since 1.7
+     */
+    @Experimental
+    record StringConcatenation(ExpressionDef left, ExpressionDef right) implements ExpressionDef {
```
dstepanov: Not sure about this one. We have `JavaIdioms::concatString` maybe we can introduce an expession but can it be a separate PR

dstepanov (changes requested):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ExpressionDef.java
andriy-dmytruk: Used that instead, as `String.concat` is not available in Kotlin

dstepanov (approved):

PR #274, opened 2025-03-20
Add support for throws clause in methods

andriy-dmytruk:


sourcegen-model/src/main/java/io/micronaut/sourcegen/model/MethodDef.java
```diff
@@ -48,6 +48,7 @@ public final class MethodDef extends AbstractElement {
     private final List<ParameterDef> parameters;
     private final List<StatementDef> statements;
     private final boolean override;
+    private final List<ClassTypeDef> throwTypes;
```
dstepanov: Considering that exceptions can be generic it should be TypeDef

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/MethodDef.java
```diff
@@ -477,6 +488,30 @@ public MethodDefBuilder addStatements(@NonNull Collection<StatementDef> newState
...
         }
 
+        /**
+         * Add throw expressions to the method.
+         *
+         * @param types The types that this method throws
+         * @return The builder
+         */
+        @NonNull
+        public MethodDefBuilder addThrows(@NonNull ClassTypeDef... types) {
```
dstepanov: Maybe lets use `throws`

sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/ByteCodeWriter.java
```diff
@@ -446,12 +446,18 @@ public void writeMethod(ClassVisitor classVisitor, @Nullable ObjectDef objectDef
         if (methodDef.isSynthetic()) {
             modifiersFlag |= ACC_SYNTHETIC;
         }
+        String[] exceptions = null;
+        if (!methodDef.getThrowTypes().isEmpty()) {
```
dstepanov, line 450: Please also add a test to `ByteCodeWriterTest`

dstepanov (changes requested):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/MethodDef.java
andriy-dmytruk: I know it would be better, but `throws` is a Java keyword

dstepanov: Right 🥹

dstepanov (approved):

PR #275, opened 2025-03-21
Support string concatenation expression

andriy-dmytruk:


dstepanov:
I think this should be implemented as `JavaIdioms.concatStringsDynamically`. Not sure I like the idea to have the dynamic call by default

dstepanov:
Or maybe to have an option with default false what kind of concat type should be used

andriy-dmytruk:
This is the default JVM behavior, which in most cases would create a `StringBuilder` underneath.

I think there should be an option to have this at least when generating Java code, as having "+" is cleaner and string builder is not always be necessary.

dstepanov:
Since when? StringBuilder was always the default. I think the dynamic stuff is only used in records

andriy-dmytruk:
This is what I found: https://www.baeldung.com/java-string-concatenation-invoke-dynamic.

Actually, it seems like I do not support constants in the same way, I can change.

dstepanov (approved):

graemerocher:
@andriy-dmytruk can you resolve the conflicts here

andriy-dmytruk:
Rebased to check if nothing will break

graemerocher (approved):

PR #276, opened 2025-03-21
More lambda improvements

dstepanov:
+ A way to resolve type variables
+ Added method type variables but they are not used now

### DIFF
--- a/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ClassTypeDef.java
+++ b/sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ClassTypeDef.java
@@ -139,7 +139,7 @@ public sealed interface ClassTypeDef extends TypeDef {
      * @return The lambda method
      * @since 1.7
      */
-    default LambdaMethodDef getLambdaMethod() {
+    default LambdaDef getLambda() {
         throw new UnsupportedOperationException("ClassTypeDef: " + getName() + " doesn't support lambdas");
     }
 
@@ -626,14 +626,14 @@ public sealed interface ClassTypeDef extends TypeDef {
     record ClassElementType(ClassElement classElement, boolean nullable) implements ClassTypeDef {
 
         @Override
-        public LambdaMethodDef getLambdaMethod() {
+        public LambdaDef getLambda() {
             List<MethodElement> abstractMethods = classElement.getEnclosedElements(
                 ElementQuery.of(MethodElement.class).onlyAbstract());
             if (abstractMethods.size() != 1) {
                 throw new IllegalArgumentException("Parent of a lambda should have exactly one " +
                     "abstract method but has " + abstractMethods.size());
             }
-            return new LambdaMethodDef(ClassTypeDef.of(classElement), MethodDef.of(abstractMethods.get(0)));
+            return new LambdaDef(ClassTypeDef.of(classElement), MethodDef.of(abstractMethods.get(0)));
         }
 
         @Override
@@ -699,7 +699,7 @@ public sealed interface ClassTypeDef extends TypeDef {
     record ClassDefType(ObjectDef objectDef, boolean nullable) implements ClassTypeDef {
 
         @Override
-        public LambdaMethodDef getLambdaMethod() {
+        public LambdaDef getLambda() {
             List<MethodDef> methods = objectDef.getMethods()
                 .stream()
                 .filter(v -> v.getModifiers().contains(Modifier.ABSTRACT))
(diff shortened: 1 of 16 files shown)


sourcegen-model/src/main/java/io/micronaut/sourcegen/model/LambdaMethodDef.java
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Implement the method as lambda.
+     *
+     * @param lambdaBuilder The lambda builder
+     * @return the lambda expression
+     * @since 1.7
+     */
+    public ExpressionDef.Lambda implement(MethodDef.MethodBodyBuilder lambdaBuilder) {
+        if (!method.getTypeVariables().isEmpty()) {
+            throw new IllegalStateException("");
```
graemerocher: can we throw a better exception message?

graemerocher (changes requested):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/LambdaMethodDef.java
dstepanov: That was a debug exception, now removed

graemerocher (approved):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/LambdaMethodDef.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Map;
+
+/**
+ * The lambda method definition.
+ *
+ * @author Denis Stepanov
+ * @since 1.7
+ */
+@Experimental
+public final class LambdaMethodDef {
```
andriy-dmytruk: Can we name this LambdaTypeDef or LambdaInterfaceDef?

It seems to contain more information than just a method.

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/LambdaMethodDef.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.util.Map;
+
+/**
+ * The lambda method definition.
+ *
+ * @author Denis Stepanov
+ * @since 1.7
+ */
+@Experimental
```
andriy-dmytruk: This seems much cleaner wit the use of this interface.

Can we document in `ExpressionDef.Lambda` that it is recommended to create it through this class, though?

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ClassTypeDef.java
```diff
@@ -130,6 +133,16 @@ default boolean isInner() {
         return false;
     }
 
+    /**
+     * Find the method that can be represented as a lambda.
+     *
+     * @return The lambda method
+     * @since 1.7
+     */
+    default LambdaMethodDef getLambdaMethod() {
```
andriy-dmytruk: Can this be `asLambdaType()` if we change the type name?

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/LambdaMethodDef.java
```diff
@@ -0,0 +1,86 @@
...
+    public ClassTypeDef getType() {
+        return type;
+    }
+
+    /**
+     * @return The lambda method
+     */
+    public MethodDef getMethod() {
+        return method;
+    }
```
andriy-dmytruk: I think we can add constants for most common lambda types, like Function, Callable, Consumer, Supplier and Predicate in the future. Wdyt?

andriy-dmytruk (approved):

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/LambdaMethodDef.java
dstepanov: LambdaDef?

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/LambdaMethodDef.java
dstepanov: Most of the use cases should not touch `new Record` style anyway

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/LambdaMethodDef.java
andriy-dmytruk: Yes, `LambdaDef` is probably the best name, thanks

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/ClassTypeDef.java
dstepanov: I feel like we want to extract/get/convert the type to lamnda somehow lambda type doesn't indicate we will select the method too

sourcegen-model/src/main/java/io/micronaut/sourcegen/model/LambdaMethodDef.java
dstepanov: Maybe in `JavaIdioms` as `JavaIdioms.functionLambda` etc not sure if this will work with Kotlin

PR #284, opened 2025-04-14
Support using builder/wither from other processors

graemerocher:
This allows avoiding having multiple rounds of processing. Also added a hook if the user needs to add logic to build method.

### DIFF
--- a/sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/BuilderAnnotationVisitor.java
+++ b/sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/BuilderAnnotationVisitor.java
@@ -110,7 +110,7 @@ public final class BuilderAnnotationVisitor implements TypeElementVisitor<Builde
             AnnotationValue<Builder> builderAnnotationValue = element.getAnnotation(Builder.class);
             ClassTypeDef elementType = ClassTypeDef.of(element);
 
-            ClassDefBuilder builder = createBuilder(elementType, builderAnnotationValue, properties, constructorParameters);
+            ClassDefBuilder builder = createBuilder(element.getPackageName(), elementType, builderAnnotationValue, properties, constructorParameters);
             ClassDef builderDef = builder.build();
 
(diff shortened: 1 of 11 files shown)


sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/WitherAnnotationVisitor.java
```diff
@@ -135,7 +107,51 @@ public void visitClass(ClassElement recordElement, VisitorContext context) {
...
+    /**
+     * Builds a wither interface for the given arguments.
+     * @param packageName The package name
+     * @param recordType The record type
+     * @param properties The properties
+     * @param parameters The parameters
+     * @param hasBuilder Is there a builder
+     * @return The interface
+     */
+    public static InterfaceDef.InterfaceDefBuilder createWither(String packageName, ClassTypeDef recordType, List<PropertyElement> properties, List<ParameterElement> parameters, boolean hasBuilder) {
```
andriy-dmytruk: In the case when original type is also generated maybe we can avoid generating the extra interface and just add the wither methods directly?

sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/BuilderAnnotationVisitor.java
```diff
@@ -145,6 +135,72 @@ public void visitClass(ClassElement element, VisitorContext context) {
...
+     *
+     * @param packageName            The package name
+     * @param elementType            The element type
+     * @param builderAnnotationValue The builder annotation value.
+     * @param properties             The properties
+     * @param constructorParameters  The constructor parameters
+     * @param buildReturnStatement   The return statement to use for building.
+     * @return A class definition builder for the builder
+     */
+    public static ClassDefBuilder createBuilder(
```
andriy-dmytruk: It could be a good idea to create a separate class for the method, like `BuilderGenerator` or `BuilderUtils` to separate the logic from the visitor.

andriy-dmytruk (approved):

dstepanov (approved):

Issue #285, opened 2025-04-15
Micronaut Sourcegen 1.7.3 broke Micronaut Platform 4.8.1 release


sdelamo:
I released 4.8.1 with sourcegen 1.7.3 and it broke starter. 

Moreover, [Micronaut SourceGen 1.7.3 should have been a minor release instead of a patch release as it seems it adds features](https://github.com/micronaut-projects/micronaut-sourcegen/releases/tag/v1.7.3). Please refer to: https://micronaut.io/micronaut-roadmap/.

> A patch GitHub Project – bug fixes / build / documentation/patch dependency updates
 

Steps to reproduce : 

Clone branch [`sourcgen` in `micronaut-starter`](https://github.com/micronaut-projects/micronaut-starter/tree/sourcgen)
```
./gradlew :starter-core:compileJava              
=========================================================
Standardized project names are disabled.
Consider enabling them with
    micronautBuild.useStandardizedProjectNames=true
in settings.gradle.

Then you will need to replace project dependencies from
   project(':foo')
to
   project(':micronaut-foo')
in your build scripts
=========================================================

> Task :starter-core:compileJava
error: cannot access GitHubApiClient$Intercepted
  bad class file: /Users/sdelamo/github/micronaut-projects/micronaut-starter/starter-core/build/classes/java/main/io/micronaut/starter/client/github/v3/GitHubApiClient$Intercepted.class
    undeclared type variable: E
    Please remove or make sure it appears in the correct subdirectory of the classpath.
/Users/sdelamo/github/micronaut-projects/micronaut-starter/starter-core/src/main/java/io/micronaut/starter/util/NameUtils.java:720: warning: [dep-ann] deprecated item is not annotated with @Deprecated
    public static boolean isGetter(String name, Class<?>[] args) {
                          ^
Note: /Users/sdelamo/github/micronaut-projects/micronaut-starter/starter-core/src/main/java/io/micronaut/starter/feature/view/React.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
1 warning

> Task :starter-core:compileJava FAILED
[Incubating] Problems report is available at: file:///Users/sdelamo/github/micronaut-projects/micronaut-starter/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
[...]

PR #288, opened 2025-04-16
Fix the erasure method in TypeDef

andriy-dmytruk:
The erasure method was broken because it assumed that `ClassTypeDef.of` would not add any type parameters.

It is fixed now by adding a `ClassTypeDef.erasure` method.

### DIFF
--- a/sourcegen-bytecode-writer/src/test/java/io/micronaut/sourcegen/bytecode/ByteCodeWriterTest.java
+++ b/sourcegen-bytecode-writer/src/test/java/io/micronaut/sourcegen/bytecode/ByteCodeWriterTest.java
@@ -2,7 +2,9 @@ package io.micronaut.sourcegen.bytecode;
 
 import io.micronaut.context.BeanResolutionContext;
 import io.micronaut.core.annotation.AnnotationClassValue;
+import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.reflect.ReflectionUtils;
+import io.micronaut.inject.ast.ClassElement;
 import io.micronaut.inject.visitor.VisitorContext;
 import io.micronaut.sourcegen.custom.visitor.GenerateLambdaVisitor;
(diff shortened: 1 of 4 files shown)


dstepanov:
I don’t think that’s enough, I’m working on more complex solution

andriy-dmytruk:
This should fix #285. I verified that micronaut core builds.

@dstepanov Now it generates introduction with erased type like:
```
class ScheduledMethodProcessor$ApplicationEventListener$scheduleTasks1$Intercepted implements ApplicationEventListener, Introduced {
    ...

    public void onApplicationEvent(Object event) {
        ...
    }
```
Before it was with incorrect generic types:
```
class ScheduledMethodProcessor$ApplicationEventListener$scheduleTasks1$Intercepted implements ApplicationEventListener<E>, Introduced {
   ...

    public void onApplicationEvent(E event) {
        ...
    }
```
I am not sure what the expected behavior is

dstepanov:
Can you please add this use-case:

```
        ClassElement predicateClass = context.getClassElement(Predicate.class).orElseThrow();
        ClassDef predicateGeneric = ClassDef.builder(element.getPackageName() + ".IfPredicateGeneric")
            .addSuperinterface(TypeDef.parameterized(ClassTypeDef.of(predicateClass), TypeDef.of(Integer.class)))
            .addMethod(MethodDef.override(predicateClass.findMethod("test").orElseThrow())
                .build((aThis, methodParameters) ->
                    methodParameters.get(0).equalsStructurally(ExpressionDef.constant(Integer.valueOf(1))).returning())
            )
            .build();

        sourceGenerator.write(predicateGeneric, context, element);
```
Add it to `GenerateIfsPredicateVisitor` + some test:

```
    @Test
    public void testGeneric() {
        IfPredicateGeneric predicate = new IfPredicateGeneric();
        assertTrue(predicate.test(1));
        assertFalse(predicate.test(2));
    }
```

dstepanov:
I see lambda is failing. I'm working on a bit different API to fix that. But maybe you can find an easier solution.

PR #304, opened 2025-05-07
Code cleanup. Fixed typos.

altro3:
Gradle 8.14.
micronaut-platform 4.8.2
Use asm version from `micronaut-core` catalog


graemerocher (approved):

build-logic/build.gradle.kts
```diff
@@ -10,4 +10,7 @@ repositories {
 dependencies {
     implementation(libs.micronaut.gradle.plugin)
     implementation(libs.sonatype.scan)
+
+    implementation(libs.managed.asm.util)
```
dstepanov: Why did you add this? Only bytecode implementation is using the ASM

altro3: @dstepanov I found the cause of the conflict. There is a version conflict here: on the one hand, a transitive dependency from micronaut core processor is used. On the other hand, the sourcegen-bytecode-writer module has a dependency on asm from the version catalog.

It is necessary to decide who is responsible for the asm core module or sourcegen . I suggest using asm from the core module.

I made 2 commits:

1. This is the solution if we want to use the asm version described in the sourcegen versions directory.

2. This is if we want to use the asm version described in the micronaut-core directory

My vision is to use the second solution

dstepanov: It should be 1. the core should not need asm dependency

PR #312, opened 2025-05-30
Fix Builder & wither with types that declare generics

graemerocher:
* For `@Wither` if the target accepts generic arguments include the generic arguments in the generated wither interface
* For `@Builder` include the generic arguments in the builder and modify `builder()` method to accept arguments.

### DIFF
--- a/sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/BuilderAnnotationVisitor.java
+++ b/sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/BuilderAnnotationVisitor.java
@@ -110,7 +110,13 @@ public final class BuilderAnnotationVisitor implements TypeElementVisitor<Builde
             AnnotationValue<Builder> builderAnnotationValue = element.getAnnotation(Builder.class);
             ClassTypeDef elementType = ClassTypeDef.of(element);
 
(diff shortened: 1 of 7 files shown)


sourcegen-generator/src/main/java/io/micronaut/sourcegen/generator/visitors/BuilderAnnotationVisitor.java
```diff
@@ -269,33 +295,66 @@ private static StatementDef mapToArrayListStatement(VariableDef.Field propertyFi
...
+        ClassTypeDef erasure = ClassTypeDef.of(builderType.getCanonicalName());
+        MethodDef.MethodDefBuilder methodBuilder = MethodDef.builder("builder")
             .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
-            .returns(builderType)
-            .addStatement(builderType.instantiate().returning())
+            .addStatement(erasure.instantiate().returning());
+        if (builderType instanceof ClassTypeDef.Parameterized parameterized) {
+            List<TypeDef> typeDefs = parameterized.typeArguments();
+            List<TypeDef> resolvedTypeDefs = new ArrayList<>();
+            int index = 1;
```
dstepanov: Not used

dstepanov (approved):

PR #319, opened 2025-06-13
Support method invocations for unbound type variables

andriy-dmytruk:
We currently support invoking type variables of form `T extends Other`, but this adds support for unbounded ones.

It also can extract a chain of bounds if for some reason those are type variables as well.

### DIFF
--- a/sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/expression/InvokeInstanceMethodExpressionWriter.java
+++ b/sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/expression/InvokeInstanceMethodExpressionWriter.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.sourcegen.bytecode.expression;
 
+import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.sourcegen.bytecode.MethodContext;
 import io.micronaut.sourcegen.bytecode.TypeUtils;
 import io.micronaut.sourcegen.model.ClassDef;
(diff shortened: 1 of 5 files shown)


sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/expression/InvokeInstanceMethodExpressionWriter.java
```diff
@@ -53,8 +54,14 @@ public void write(GeneratorAdapter generatorAdapter, MethodContext context) {
         Type methodOwnerType = TypeUtils.getType(instanceType, context.objectDef());
         MethodDef methodDef = invokeInstanceMethod.method();
         Method method = new Method(methodDef.getName(), TypeUtils.getMethodDescriptor(context.objectDef(), methodDef));
-        if (instanceType instanceof TypeDef.TypeVariable typeVariable) {
-            instanceType = typeVariable.bounds().get(0);
+        while (instanceType instanceof TypeDef.TypeVariable typeVariable) {
```
andriy-dmytruk, line 57: It fixes https://github.com/micronaut-projects/micronaut-security/issues/1981.

What is strange is that the type being called in security is actually
```java
TypeVariable[
  name=V, 
  bounds=[
     TypeVariable[
       name=V,
       bounds=[
         ClassElementType[
           classElement=java.lang.Object, 
           nullable=false
         ]
       ], 
       nullable=false
     ], 
     ClassElementType[
       classElement=java.lang.Object, 
       nullable=false
     ]
  ], 
  nullable=false
]
```

So it is like saying `V extends (V extends Object)`

PR #367, opened 2025-12-24
Fix Wither builder type resolution for nested records fix #366

### LINKED ISSUE #366: Wither generator resolves wrong builder FQN for nested records when combined with @Builder
unable to use @Builder and @Wither on nested records - wither generated with wrong reference to builder.
step to reproduce:
Create class, like

```
public final class Foo {
    @Wither
    @Builder
    protected record Bar(String siteName, String buildingName, String roomNumber)
        implements FooBarWither {}
}
```
during the build invalid interface is generated:
```
public interface Foo$BarWither {
  default Foo.Bar$NodeGroupInfoBuilder with() {
     return new demo.test.Foo.Foo$BarBuilder(this.siteName(), this.buildingName(), this.roomNumber());
  }
}
```
in generated code classes is not nested under Foo and should be:
```
public interface Foo$BarWither {
  default Foo.Bar$NodeGroupInfoBuilder with() {
     return new demo.test.Foo$BarBuilder(this.siteName(), this.buildingName(), this.roomNumber());
  }
}
```
logs:
error: cannot find symbol
    default Foo.Foo$BarBuilder with() {
cannot find symbol
    default Foo.Bar with(Consumer<Foo.Foo$BarBuilder> consumer) {
                                     ^
    symbol:   class Foo$BarBuilder
    location: class Foo

### Version

4.9.1

araysse:
Root Cause

When using @Wither and @Builder annotations together on nested records (records defined inside a class), the generated wither interface incorrectly references the builder class name, causing compilation errors.

Approach

- Derive the builder’s package from the binary name (the string with $ for inner classes), not the canonical name.

- Compute the package as the substring of recordType.getName() up to the last '.'.

- Build the builder simple name using the binary part if the record is inner:

  - If inner: FooBarBuilder
  - Else: FooBuilder

- Compose the final FQN as:
  - builderClassName = pkg.isEmpty() ? builderSimpleName : pkg + "." + builderSimpleName

- Preserve generics by using TypeDef.parameterized when the record type is parameterized.

Verification

- Reproduced failure with a nested record using both annotations; after applying the change, build succeeds.

Closes: #366

### DIFF
--- a/sourcegen-generator-kotlin/build.gradle.kts
+++ b/sourcegen-generator-kotlin/build.gradle.kts
@@ -16,9 +16,3 @@ dependencies {
 tasks.test {
     develocity.predictiveTestSelection.enabled = false
 }
-
-kotlin {
-    compilerOptions {
-        jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17
-    }
-}
--- a/sourcegen-generator-kotlin/build.gradle.kts
+++ b/sourcegen-generator-kotlin/build.gradle.kts
@@ -16,3 +16,9 @@ dependencies {
 tasks.test {
     develocity.predictiveTestSelection.enabled = false
 }
+
+kotlin {
+    compilerOptions {
+        jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17
+    }
+}
--- a/sourcegen-generator-kotlin/build.gradle.kts
+++ b/sourcegen-generator-kotlin/build.gradle.kts
@@ -3,6 +3,12 @@ plugins {
     alias(mn.plugins.kotlin.jvm)
 }
 
+kotlin {
+    compilerOptions {
+        jvmTarget.set(org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17)
+    }
+}
+
 dependencies {
     implementation(projects.sourcegenGenerator)
     implementation(libs.managed.kotlinpoet)
--- a/sourcegen-generator/src/test/groovy/io/micronaut/sourcegen/generator/visitors/WitherAnnotationVisitorSpec.groovy
+++ b/sourcegen-generator/src/test/groovy/io/micronaut/sourcegen/generator/visitors/WitherAnnotationVisitorSpec.groovy
@@ -1,0 +1,67 @@
+package io.micronaut.sourcegen.generator.visitors
+
+import io.micronaut.annotation.processing.test.AbstractTypeElementSpec
+
+class WitherAnnotationVisitorSpec extends AbstractTypeElementSpec {
+
+    void "inner record with @Wither and @Builder resolves builder type in correct package and compiles"() {
+        given:
+        // Reproduction: top-level class Foo with protected inner record Bar using @Wither and @Builder
+        // The wither must reference the correct builder type for the inner record without duplicating the package.
+        var classLoader = buildClassLoader("demo.test.Foo", """
+        package demo.test;
+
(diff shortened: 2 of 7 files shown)


dstepanov:
Please add a test that is reproducing this problem

dstepanov:
Why did you add JVM_17?

sdelamo (changes requested):
The pull request misses tests. Is there an issue this PR aims to fix?

sourcegen-generator-kotlin/build.gradle.kts
```diff
@@ -16,3 +16,9 @@ dependencies {
 tasks.test {
     develocity.predictiveTestSelection.enabled = false
 }
+
+kotlin {
+    compilerOptions {
+        jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17
+    }
+}
```
sdelamo:
```suggestion
```

test-suite-kotlin/build.gradle.kts
```diff
@@ -25,3 +25,8 @@ dependencies {
 //kotlin {
 //    kotlinDaemonJvmArgs = listOf("-Xdebug","-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y")
 //}
+kotlin {
+    compilerOptions {
+        jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17
+    }
+}
```
sdelamo:
```suggestion
```

sdelamo (changes requested):

araysse:
> Why did you add JVM_17?

i just need to make the build pass because i was have
Inconsistent JVM-target compatibility detected for tasks 'compileJava' (17) and 'compileKotlin' (21).
this pr fix the issue [#366](https://github.com/micronaut-projects/micronaut-sourcegen/issues/366)

sourcegen-generator/src/test/groovy/io/micronaut/sourcegen/generator/visitors/WitherAnnotationVisitorSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+        var classLoader = buildClassLoader("demo.test.Foo", """
+        package demo.test;
+
+        import io.micronaut.sourcegen.annotations.Wither;
+        import io.micronaut.sourcegen.annotations.Builder;
+
+        public class Foo {
+            @Wither
+            @Builder
+            protected record Bar(String a, String b, String c) implements Foo\$BarWither {
```
sdelamo: users should not write code coupled to internal classes such as `Foo\$BarWither`

dstepanov: How do you add Wither then? Why does it add `$` for the inner classes?

araysse: Add Wither by annotating the record with @Wither (and @Builder if you want a builder). The annotation processor generates the companion Wither interface; user code should not implement it manually.

dstepanov: Did you answer my question with AI?
Withers should be used by adding it as implements interface like it was before so the previous example was correct. It's just I don't think we should generate `$`, it should be `implements Foo.BarWither`

dstepanov (approved):

Issue #387, opened 2026-02-19
@Builder annotation doesn't generate defaults for primitive types


leo-orcl:
### Expected Behavior

Consider the following test:

```java
import static org.junit.jupiter.api.Assertions.assertEquals;

import org.junit.jupiter.api.Test;

public class BuilderTest {

    @io.micronaut.sourcegen.annotations.Builder
    public record MicronautRecord(int primitive) {}

    @lombok.Builder
    public record LombokRecord(int primitive) {}

    @Test
    public void testMicronaut() {
        MicronautRecord record = BuilderTestMicronautRecordBuilder.builder().build();
        assertEquals(0, record.primitive());
    }

    @Test
    public void testLombok() {
        LombokRecord record = LombokRecord.builder().build();
        assertEquals(0, record.primitive());
    }
}
```

I expect the behavior of builders produced with Micronaut to be similar to the ones produced by Lombok, i.e., both tests should pass.

### Actual Behaviour

Instead, it fails with:

```
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because "this.primitive" is null

  at BuilderTestMicronautRecordBuilder.build(BuilderTestMicronautRecordBuilder.java:29)
  at BuilderTest.testMicronaut(BuilderTest.java:17)
  ...
```

### Steps To Reproduce

Just compile and run the test above.

### Environment Information

Operating System: macOS Tahoe 26.2
JDK: 21.0.9-graal

### Version

1.8.4

ZARAYACH:
This should be fixed in [v1.8.5](https://github.com/micronaut-projects/micronaut-sourcegen/releases/tag/v1.8.5). this issue can be closed.

PR #392, opened 2026-02-26
Fixing primitive types default values in builder generated classes

ZARAYACH:
Hello, 
This pr fixes the bug mentioned in the issue #387 the problem was that it was using the default constructor and when building the object without initializing the primitive type it throws a null pointer exception cause you can't have null in primitive variables , the fix is adding inline if statement in the constructor parameters for the primitives types , and it wasn't catch by the tests cause it initialized every int variable to a value of 10 and there were no other primitive fields


dstepanov (approved):

dstepanov:
Pls fix checkstyle. Not sure if copilot review make sense

PR #404, opened 2026-03-11
Fix sonar and other warnings

dstepanov:


dstepanov:
Sonar still failing because the code coverage is not properly calculated

graemerocher (approved):

PR #454, opened 2026-06-28
Support annotation class generation in the Kotlin and bytecode generators

### LINKED ISSUE #283: Support for annotation classes generation in Kotlin & bytecode
### Feature description

Support for annotation classes was added in #282.

For consistency it should probably be added to Kotlin & bytecode generators.
I don't actually require those, but adding this issue to keep track of the differences between generators.

GabrielBBaldez:
## What

Adds annotation-class generation (`AnnotationObjectDef`) to the Kotlin (`KotlinPoetSourceGenerator`) and bytecode (`ByteCodeWriter`) generators, so all three backends can emit annotation types — the Java generator already does (#282).

Fixes #283

## How

- **Kotlin**: emits a Kotlin `annotation class`, mapping members to primary-constructor `val` parameters with their defaults; primitive array members use the specialized Kotlin array types (`IntArray`, `FloatArray`).
- **Bytecode**: emits an `ACC_ANNOTATION | ACC_INTERFACE | ACC_ABSTRACT` interface implementing `java.lang.annotation.Annotation`, one accessor method per member, with `visitAnnotationDefault` for defaults.
- Array-valued meta-annotation members (e.g. `@Target`) are supplied as collections in the shared test fixture, so both generators render them through their existing collection handling — no reflection or type inference at generation time.

## Tested

- **Kotlin**: the generated annotation is compiled and applied in `test-suite-kotlin` (KSP round-trip), plus a golden write test.
- **Bytecode**: the generated class is loaded and verified reflectively — `@Retention`/`@Target`, member return types and defaults — and validated with ASM's `CheckClassAdapter`.

## Note

The Kotlin generator omits a member whose default value is a *nested annotation* typed as a Java annotation (the fixture's `inner`). The Kotlin compiler aborts with an internal error when a Java annotation is emitted in default-value position — the member type and ordinary usages compile fine, only that default cannot be expressed. The Java and bytecode generators emit it normally.

### DIFF
--- a/sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/ByteCodeWriter.java
+++ b/sourcegen-bytecode-writer/src/main/java/io/micronaut/sourcegen/bytecode/ByteCodeWriter.java
@@ -19,6 +19,7 @@ import org.jspecify.annotations.Nullable;
 import io.micronaut.core.naming.NameUtils;
 import io.micronaut.sourcegen.bytecode.statement.StatementWriter;
 import io.micronaut.sourcegen.model.AnnotationDef;
+import io.micronaut.sourcegen.model.AnnotationObjectDef;
 import io.micronaut.sourcegen.model.ClassDef;
 import io.micronaut.sourcegen.model.ClassTypeDef;
 import io.micronaut.sourcegen.model.EnumDef;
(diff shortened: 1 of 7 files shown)


dstepanov:
Thanks for this, and sorry it sat so long.

Closing as superseded by #502. The Kotlin half landed separately in #487, which also handles the nested-annotation default this PR had to leave out. Since then a second bytecode backend arrived (#497), so #502 does the bytecode work in both backends against `2.2.x`.

PR #496, opened 2026-09-01
Support staged builders

### LINKED ISSUE #198: Support staged builders
### Feature description

See http://immutables.github.io/immutable.html#staged-builder

dstepanov:
Adds `@StagedBuilder`, which generates a builder that assigns the required properties in stages, as described in https://immutables.github.io/immutable.html#staged-builder

Fixes #198

## The generated shape

For:

```java
@StagedBuilder
public record Employee(String name, int age, boolean employed, @Nullable String nickname) {
}
```

the generated `EmployeeStagedBuilder` is:

```java
public final class EmployeeStagedBuilder {
    public static NameBuildStage builder() { ... }

    public interface NameBuildStage {
        AgeBuildStage name(String name);
    }
    public interface AgeBuildStage {
        EmployedBuildStage age(int age);
    }
    public interface EmployedBuildStage {
        BuildFinal employed(boolean employed);
    }
    public interface BuildFinal {
        BuildFinal nickname(String nickname);

        Employee build();
    }

    public static final class Builder implements NameBuildStage, AgeBuildStage, EmployedBuildStage, BuildFinal { ... }
}
```

Every required property gets a stage interface declaring the single method that assigns it and returns the next stage, so `build()` is only reachable once all of them are assigned:

```java
var employee = EmployeeStagedBuilder.builder()
    .name("Billy Bounce")
    .age(33)
    .employed(false)
    .nickname("Billy")
    .build();
```

A property is required unless the builder can leave it unassigned with a defined value: it is nullable, it declares a `@Bindable(defaultValue = "...")` default value, or it accumulates values with `@Singular`. Those are declared on `BuildFinal` and can be assigned in any order, or not at all.

The generated type holds the stages and the `builder()` method, and the builder itself is the nested `Builder` class implementing every stage — a class cannot implement an interface nested in itself (javac reports `cyclic inheritance`), so the two cannot be the same type. This is also the shape Immutables generates. Generic types are supported: a nested interface cannot inherit the type variables of its enclosing type, so every stage redeclares them.

## Backend fixes needed for this shape

- `KotlinPoetSourceGenerator` rejected the `static` modifier of a nested type definition (`Not supported modifier: static`), which a Kotlin nested type is by default. It is now stripped from type modifiers, the way it already was for methods and fields.
- The bytecode writer erased a method type variable used as a type argument — `builder()` was written with the signature `()LContentBuildStage<Object>;` instead of `()LContentBuildStage<TS>;` — because the recursive calls of the signature writer did not pass the method being written. Covered by the generic `Cargo` test in `test-suite-bytecode`, which does not compile without the fix.

## Refactoring

`BuilderAnnotationVisitor` now returns the methods it creates for a property and accepts an explicit return type and an override flag, so the staged builder can reuse the property, singular, constructor and build method generation. `@Builder` and `@SuperBuilder` output is unchanged.

## Tests and docs

- `Employee` (required + nullable), `Project` (`@Bindable` default + `@Singular`) and `Cargo` (generic) models with tests in `test-suite-java` and `test-suite-bytecode`, and a data class with a test in `test-suite-kotlin`.
- New `Staged Builder` guide page, plus the annotations table and `toc.yml`.

Verified with `./gradlew test checkstyleMain spotlessCheck docs` and `./gradlew :test-suite-java:test :test-suite-bytecode:test :test-suite-kotlin:test --rerun-tasks`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FgEP1bRJtu22ipXnFQbLtQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01FgEP1bRJtu22ipXnFQbLtQ)_

### DIFF
--- a/src/main/docs/guide/annotations/stagedBuilder.adoc
+++ b/src/main/docs/guide/annotations/stagedBuilder.adoc
@@ -1,0 +1,51 @@
+The `@StagedBuilder` annotation generates a https://immutables.github.io/immutable.html#staged-builder[staged builder]: a builder that assigns the required properties in stages, so that forgetting one of them is a compilation error instead of a missing value at runtime.
+
+If you annotate a Java Record with `@StagedBuilder`, an `EmployeeStagedBuilder` class is generated at compilation-time.
+
+[source,java]
+----
+include::test-suite-java/src/main/java/io/micronaut/sourcegen/example/Employee.java[tags=clazz,indent=0]
+----
+
+Every required property gets its own stage interface, which declares the single method that assigns that property and returns the next stage. The last of them returns `BuildFinal`, the only stage that declares the `build` method:
+
+[source,java]
+----
+public final class EmployeeStagedBuilder {
+    public static NameBuildStage builder() { ... }
+
+    public interface NameBuildStage {
+        AgeBuildStage name(String name);
+    }
+    public interface AgeBuildStage {
+        EmployedBuildStage age(int age);
+    }
+    public interface EmployedBuildStage {
+        BuildFinal employed(boolean employed);
+    }
+    public interface BuildFinal {
+        BuildFinal nickname(String nickname);
+
+        Employee build();
+    }
+
+    public static final class Builder implements NameBuildStage, AgeBuildStage, EmployedBuildStage, BuildFinal { ... }
+}
+----
+
+You can use the builder to create an `Employee` instance, assigning the required properties in the order they are declared in:
+
+[source,java]
+----
+include::test-suite-java/src/test/java/io/micronaut/sourcegen/example/EmployeeStagedBuilderTest.java[tags=test,indent=0]
+----
+
+A property is required unless the builder can leave it unassigned, which is the case for:
+
+- a nullable property, which keeps its `null`
+- a property with a `@Bindable(defaultValue = "...")` default value, which keeps that value
+- a property annotated with api:sourcegen.annotations.Singular[], which accumulates into an empty collection
+
+Those properties are declared on the `BuildFinal` stage instead and can be assigned in any order, or not at all.
+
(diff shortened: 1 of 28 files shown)


src/main/docs/guide/annotations/stagedBuilder.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+    public static final class Builder implements NameBuildStage, AgeBuildStage, EmployedBuildStage, BuildFinal { ... }
+}
+----
+
+You can use the builder to create an `Employee` instance, assigning the required properties in the order they are declared in:
+
+[source,java]
+----
+include::test-suite-java/src/test/java/io/micronaut/sourcegen/example/EmployeeStagedBuilderTest.java[tags=test,indent=0]
```
graemerocher: can we replace these with snippet: that include the Kotlin samples?

graemerocher (changes requested):

src/main/docs/guide/annotations/stagedBuilder.adoc
dstepanov: Done in 49c360f — both blocks are `snippet::` now, so the guide shows the Java and the Kotlin sample:

```
snippet::io.micronaut.sourcegen.example.Employee[project-base="test-suite", source="main", tags="clazz", indent=0]
snippet::io.micronaut.sourcegen.example.EmployeeStagedBuilderTest[project-base="test-suite", tags="test", indent=0]
```

The Kotlin example and its test needed the tags to include. It also no longer annotates the nullable property with `@Nullable`: a Kotlin nullable type is one already, which the staged builder reads, so the annotation only made the sample noisier — the Kotlin tests still put `nickname` on the final stage without it.

Built the guide and checked the rendered page carries a Java and a Kotlin tab for both snippets.

---
_Generated by [Claude Code](https://claude.ai/code)_

graemerocher (approved):

PR #497, opened 2026-09-02
Add JDK ClassFile bytecode backend

graemerocher:
## Summary

Adds an opt-in JDK 25 ClassFile bytecode backend that emits Java 17-compatible classfiles, alongside the existing ASM backend.

## Included

- Backend-neutral bytecode writer core utilities.
- Direct JDK ClassFile writer with stack maps, verification, records, annotations, invokedynamic, control flow, bridges, and Sourcegen-aware hierarchy resolution.
- JDK generator service and reusable JDK bytecode integration runner.
- Shared generator traversal, BOM/settings wiring, and documentation.
- Expanded JDK parity coverage requiring direct ClassFile lowering for writer tests.

## Verification

- ASM writer: 149 tests passed.
- JDK writer: 22 tests passed.
- JDK integration suite: 128 tests, 14 skipped, 0 failures.
- `spotlessCheck`, `japiCmp`, root `check`, and `git diff --check` passed.
- JDK writer and generator runtime dependency graphs contain no ASM artifacts.

This is a draft for implementation and API review.


dstepanov (approved):
Reviewed the backend end to end and validated it against micronaut-core. Rebased onto the latest `2.2.x` and pushed.

## Verification

All on JDK 25.

| What | Result |
|---|---|
| `./gradlew check` (all modules) | 1446 tests, 0 failures |
| micronaut-core 5.2.x compiled with this backend | every module, no definition falling back to javac or to source |
| micronaut-core inject / context / test-suite | 8319 tests, 0 failures |
| micronaut-serialization | 6 failures, byte-for-byte the same set as the untouched baseline |
| micronaut-data | 2 failures, same set as the untouched baseline |

micronaut-core was patched locally so `ByteCodeWriterUtils` calls the JDK writer instead of ASM, then compiled and tested. The downstream failures were checked against pristine micronaut-core resolving the released sourcegen 2.1.0 and reproduce identically there, so none is attributable to this PR: the data ones are a timezone-dependent `LocalDate` to `java.util.Date` conversion, and the serde ones are Jackson Databind behaviour tests.

## What the validation turned up

Driving a real workload through the writer found the issues that unit tests had not, all now fixed and covered:

- `super` was rejected as an invocation receiver, which declined every bean definition.
- `isDefault` emitted `invokespecial` for any call to a default method, which is invalid unless the interface is a direct superinterface.
- A cast chain kept every cast, so a value cast to a primitive and back to its wrapper was unboxed and reboxed, turning a null Kotlin default argument into a `NullPointerException`.
- A `super(...)` written in the deprecated instance-invocation form was read as `this(...)` delegation, so field initializers were skipped and final fields stayed null.
- Multi-dimensional arrays allocated one dimension; array constants were unsupported.
- Keys sharing one case body each got their own copy, and switches always used `lookupswitch`, which pushed a generated dispatch method past the 64KB limit.

One of these was not in the new backend at all: `bytecode-writer-core` was an `implementation` dependency of the ASM writer, so it was missing from the compile classpath of consumers that run the writer from there. Micronaut's Groovy AST transform does exactly that, and every Groovy build failed with `NoClassDefFoundError`. Both writers now expose it as `api`.

## Rebase

Rebased onto `25b84a9e`. #498 changed `ByteCodeWriter` and `SignatureWriterUtils`, the two files this PR extracts into the shared core, so I kept upstream's versions of both: the ASM backend keeps #498's signature fixes and retention handling unchanged, and `BridgeResolver`, `EnumGenUtils` and `TypeUtils` still delegate to the core.

## Follow-ups, none of them blocking

The JDK backend is `@Experimental` and opt-in, so these are worth filing rather than holding the PR for:

1. **Generic signatures are simpler than ASM's.** After #498 the ASM writer emits real wildcard type arguments (`*`, `+`, `-`), honours every bound of a type variable, and recurses into array components when deciding whether a signature is needed. The core's implementation writes a wildcard as `Object`, uses only the first bound, and does not look through arrays. Generated classes verify and run, but their generic metadata is poorer than the ASM backend's for the same model.
2. **`RetentionPolicy` is not honoured.** #498 made the ASM writer skip `SOURCE` annotations and write `CLASS` ones invisible. The JDK writer still writes every annotation as runtime-visible.
3. **The source fallback is still reachable** for switch yield cases and a few other constructs. It now warns when it fires, which is the important part, but the remaining gaps are worth closing so the fallback can eventually be dropped.

Approving: the backend does what it says, it is opt-in, it does not disturb the ASM path, and it compiles and passes the whole of micronaut-core.

PR #504, opened 2026-09-04
Write class-valued annotation members with the class tag in the JDK backend

dstepanov:
## What

The JDK ClassFile backend (`sourcegen-bytecode-writer-jdk`) dropped a `Class`-valued annotation member when it copied an annotation. Given a fixture annotated `@Simple(name = "A", age = 12, color = Color.WHITE, type = String.class)`, the generated class came back with `stringValue("name")`, `intValue("age")` and `enumValue("color", Color.class)` all resolving, but `classValue("type")` empty. The ASM backend and the javac backend both handled it correctly.

## Why

A class-valued member is modelled as a `VariableDef.StaticField` named `"class"` whose type is `TypeDef.CLASS` and whose owner is the referenced type — i.e. `SomeType.class`. The JDK backend's `StaticField` branch in `ByteCodeWriter.toAnnotationValue` assumed every static field was an enum constant, so it emitted `AnnotationValue.ofEnum(...)` and wrote the member with the `e` tag rather than the `c` tag. The reader then found no class value under that name.

The fix special-cases that shape and emits `AnnotationValue.ofClass` with the owner type's descriptor, mirroring what the ASM backend already does in `sourcegen-bytecode-writer`'s `ByteCodeWriter#visitAnnotation`. The `ClassTypeDef` / `Class<?>` branches further down were already correct — they were simply never reached for this shape of value.

## Test

`test-suite-java`'s `HeronTest` already asserted `classValue("type")`, which is why the javac backend was known good. There was no bytecode-side equivalent, so neither bytecode backend was held to it. This adds `HeronTest` to `test-suite-bytecode`, whose test sources `test-suite-bytecode-jdk` also compiles — so the assertion now runs against the ASM backend and the JDK backend from one source.

Verified the test catches the bug: with the six-line fix reverted, `:test-suite-bytecode-jdk:test` fails on `copiesEveryMemberOfAnAnnotation()` with the original `java.util.NoSuchElementException: No value present`.

- `:test-suite-bytecode-jdk:test` — 139 pass
- `:test-suite-bytecode:test` — 139 pass
- `:test-suite-java:test` — pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sourcegen-bytecode-writer-jdk/src/main/java/io/micronaut/sourcegen/bytecode/jdk/ByteCodeWriter.java
+++ b/sourcegen-bytecode-writer-jdk/src/main/java/io/micronaut/sourcegen/bytecode/jdk/ByteCodeWriter.java
@@ -542,6 +542,12 @@ public final class ByteCodeWriter {
             return toAnnotationValue(constantValue);
         }
         if (value instanceof VariableDef.StaticField field) {
+            if (field.name().equals("class") && field.type().equals(TypeDef.CLASS)) {
+                // A class-valued member is modelled as `SomeType.class`, not as an enum constant
+                return AnnotationValue.ofClass(ClassDesc.ofDescriptor(
+                    io.micronaut.sourcegen.bytecode.core.TypeUtils.getDescriptor(field.ownerType(), null)
+                ));
+            }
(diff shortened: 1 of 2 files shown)


graemerocher:
can we merge the other TCK first, since it makes it difficult to review these PRs

